PR #1, opened 2018-03-05
Create gem to distribute the tech docs template

tijmenb:
This converts the tech-docs-template repo into a gem. It is a fork of the original https://github.com/alphagov/tech-docs-template, which will be updated to use this gem. This will mean the 2 repos will have a shared git history.

What happens in this commit?

- Adds a `.gemspec` so that we'll be able to release this as a gem
- Adds a `example` directory, which is an example of a documentation site. It can be used in day-to-day development and is used to build an integration test.
- Integrate the [table_of_contents](https://github.com/alphagov/table_of_contents) gem, which is only used for these tech docs. Because the gem was never released to Rubygems we can't make it a dependency of this gem (even if we wanted to).
- Add `TechDocs.configure(self)` which sites use to configure itself. I've done some experimentation to make this a proper Middleman extension but that seemed to be a pretty unreliable way of configuration.

## Testing

You can see it in action here:

- https://github.com/alphagov/govuk-content-api-docs/pull/45 updates the content API docs

Part of https://github.com/alphagov/tech-docs-template/issues/126.


tijmenb:
> Do still think you should prefix it with govuk if it's going to rubygems.

I've renamed it to `govuk_tech_docs` everywhere.

kevindew:
yay have a big ✅ then.

PR #19, opened 2018-04-25
Determine active page by considering parent too

quis:
Currently the logic that works out which page to highlight as ‘active’ in the top navigation is quite naïve (only compares URLs directly).

This commit makes it a bit smarter so that a page can define its `parent`, then the navigation will highlight the matching ‘parent’ navigation item.

Adapted from: https://github.com/alphagov/govuk-developer-docs/commit/7e5ac5144d87aebf28ca98297b4250c307416c8e

### DIFF
--- a/lib/govuk_tech_docs.rb
+++ b/lib/govuk_tech_docs.rb
@@ -70,6 +70,14 @@ module GovukTechDocs
       def format_date(date)
         date.strftime('%-e %B %Y')
       end
+
+      def active_page(page_path)
+        [
+          page_path == "/" && current_page.path == "index.html",
+          ("/" + current_page.path) == page_path,
+          current_page.data.parent != nil && ("/" + current_page.data.parent.to_s) == page_path,
+        ].any?
+      end
     end
 
     context.page '/*.xml', layout: false
--- a/docs/frontmatter.md
+++ b/docs/frontmatter.md
@@ -110,4 +110,14 @@ title: My beautiful page
 ---
 ```
 
+## `parent`
+
+The page that should be highlighted as ‘active’ in the navigation.
+
+```yaml
+---
+parent: shaving-yaks.html
+---
+```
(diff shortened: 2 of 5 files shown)


lib/govuk_tech_docs/helpers.rb
```diff
@@ -0,0 +1,17 @@
...
+      @current_page = current_page
+    end
+
+    def from_path?(page_path)
+      ("/" + @current_page.path) == page_path || @current_page.data.parent == page_path
+    end
+
+  private
+
+    attr_reader :config, :current_page
```
tijmenb: I think these aren't necessary, since you're using `@current_page` page above.

lib/govuk_tech_docs.rb
```diff
@@ -70,6 +71,10 @@ def current_page_review
       def format_date(date)
         date.strftime('%-e %B %Y')
       end
+
+      def active_page
```
tijmenb: Have you considered making moving the `from_path?` logic into in here?

quis:
OK, I’ve managed to make the code much simpler. Not sure how to test it though – I don’t think I can copy the example you gave now that the implementation isn’t wrapped up in a class.

lib/govuk_tech_docs.rb
```diff
@@ -70,6 +70,10 @@ def current_page_review
       def format_date(date)
         date.strftime('%-e %B %Y')
       end
+
+      def active_page(page_path)
+        ("/" + current_page.path) == page_path || current_page.data.parent == page_path
```
tijmenb: I think this might not work for the homepage, because `page_path` is `/` and `current_page.path` is `index.html`.

quis:
@tijmenb I have added:
- tests 
- documentation
- a whole bunch of extra complexity to make it actually work 😞

PR #27, opened 2018-06-12
GTD-3 Multipage Navigation

### LINKED ISSUE #128: Navigation is broken when deploying on a path
The left-hand side navigation doesn't look for the defined host (`host: https://alphagov.github.io/api-catalogue`) generating absolute urls from the root `/` causing broken links when deployed on GitHub pages (or to a specific folder).

Note: this only happens in a multi-page config.

lewisnyman:
Fixes alphagov/tech-docs-template#128.

This adds multiple page navigation support for the sidebar navigation. When you create additional .html.md files, they'll be included in the sidebar, along with the headings on that page. These are collapsed by default. 

kooctsbmq9

I've got this running against a fork of the paas content here: ConvivioTeam/paas-tech-docs/tree/GTD-14-prototype-multipage-navigation

Currently it only displays the top level pages, it doesn't' support nested pages within pages. I'm eager to get this used as soon as possible so we can figure out where we need more flexibility. This is my first PR against this project so nitpicking is welcome!


docs/frontmatter.md
```diff
@@ -110,6 +110,17 @@ title: My beautiful page
 ---
 ```
 
+## `weight`
+
+Affects the order a page is displayed in the sidebar navigation tree. Lower
```
tijmenb: This confused me a bit because a higher weight in search results often means that things will end up higher in the results. What would you think of flipping the logic and naming this `navigation_priority`?

Big 👍 for documenting the option.

lib/source/layouts/layout.erb
```diff
@@ -5,10 +5,37 @@ wrap_layout :core do
   content_for(:toc_module, "in-page-navigation")
 
   content_for :sidebar do
-    table_of_contents(
-      html,
-      max_level: config[:tech_docs][:max_toc_heading_level]
-    )
+
+    # Only use html files
```
tijmenb: Perhaps this can be extracted into a helper, so we can test it separately.

docs/frontmatter.md
lewisnyman: Sure, that seems more explicit. I'll ask @jonathanglassman for his thoughts.

jonathanglassman: I'm happy with either. "Weight = heavier sinks to the bottom" made sense to me, but navigation priority also makes sense. I would teach it as "smaller numbers = higher up the TOC" so as long as people don't think that higher numbers mean a higher priority in the TOC, we should be ok. If I had to make a choice I would keep it as is.

lewisnyman: Yeah I see your point, it would be nice to have something that just makes sense without an explanation. It's hard to come up with a scale that's intuitive.

Jon and I had a quick chat this morning, and decided to stick with weight for now and he really connected with the analogy. However, we want to push this out wider to other writers and see if they get confused by this.

lewisnyman:
Ok, I've written some tests, I copied a lot from the meta tags unit tests.

> I wonder if this needs to be optional behaviour. It works a bit oddly with the current setup of the GOV.UK docs. If it was a configuration option we could merge this sooner and test it without compatibility issues.
I think that we need to work towards supporting more complex navigation structures like gov.uk docs. I think that might be the most complex use case.

I was thinking we could release this on a major version, then some sites can choose to hold back until it's ready.

tijmenb:
@lewisnyman I would like to release as few breaking versions of this as possible. Existing sites won't always have a developer available so upgrading should be made as easy as possible. Also I'm not sure this particular pattern will work for GOV.UK, so I'd like to keep the option of using the existing functionality.

docs/frontmatter.md
MatMoore: I haven't had time to look at the PR as a whole yet, but just wanted to "weigh in" on this discussion... 

I would interpret weight in the same way as @tijmenb as in more weight = more important = more prominent.

I would call it something like `navigation_order`.

MatMoore:
I have a few questions that aren't really related to the PR, but are more about the feature in general. Happy to discuss offline if this isn't the best place for it.

- Do we have any usability testing planned for the multi-page pattern?
- Do we have any assumptions/recommendations about when to use this vs a single page layout?
- Do we have strong opinions about when you would put something in the horizontal nav at the top vs the sidebar nav? For example I'm not really familiar with the PaaS docs, but what makes the "about" and "features" pages not part of the documentation itself?
- Do we have a list of sites we plan on moving to the multi page pattern (or is it all of them?)

I think it's fine if we say this is an optional feature that not all sites will use, but I'm wary about putting too much choice in the hands of the whoever's consuming the gem by making things like this opt-in. IMO we could push for more consistency in how it's used.

There are certain kinds of documentation (like API docs) that could follow a very standard approach and we already have a lot of guidance about the kinds of things we expect to be there.

If someone's already thought about the general information architecture for that use case, then I would prefer for the gem to encourage that approach for users who are spinning up new sites. Users of the gem should be able to set something up quickly without doing a lot of design work, and we should avoid solving similar problems differently across different teams.

jonathanglassman:
@MatMoore responses to your questions:
1. We have some usability testing planned for tomorrow - tech writers plus Arnau from Registers will be testing out the functionality and providing feedback.
2. No formal assumptions yet; I believe it should be looked at on a case by case basis and agreed between the tech writer and their team, but happy to discuss what the right approach is to take.
3. The documentation is usually for technical audience i.e. the dev trying to accomplish something like integrate their service. The about and features sections cover content that does not fit into this category. Generally it is "product" information.
4. See answer to 2. Personally I believe that Pay and PaaS need to go multipage, and potentially Verify as well, due to amount of content. However, again I believe this should be discussed and agreed with tech writers and teams.

Agree that we should provide guidance on how it should be used. The standard approaches that we use for certain types of content like API docs are still compatible with this approach.

lewisnyman:
Here's a visual update:
screenshot 2018-06-19 10 54 18

lewisnyman:
Screenshots on iPhone 5s and iPhone X
screenshot 2018-06-19 11 40 57
screenshot 2018-06-19 11 40 54

lewisnyman:
I've added a boolean option: `multipage_nav` that enables this new functionality. It defaults to false. When this is set to false, it will only render the navigation tree of the page you area currently on, it doesn't stop you creating multiple pages (see gov.uk dev docs)

This option does no affect the collapsible navigation UI, at least not yet. It feels like there might be some value in having this for single page sites.

lewisnyman:
We ran a testing workshop on the new multipage functionality. Here's some of the feedback related to the comment above:

We had some feedback that collapsible navigation would be useful on sites that wanted to stick with a single page structure eg Registers. It sounds like it's a good idea to keep that functionality separate to the the multipage_nav option.

Does the collapsible functionality needs it's own option to enable? I haven't seen a situation where that is the case yet.

jonathanglassman:
We also had a discussion on the weight issue. Although it seemed to make intuitive sense, there are potential complications arising from implementing multipage functionality with groupings and "nested" subpages. @lewisnyman will build this subpage functionality and then we can test if the weight method works, or if we need to implement another method, such as a TOC page that recreates the page hierarchy in a single markdown file.

lewisnyman:
> Does the collapsible functionality needs it's own option to enable? I haven't seen a situation where that is the case yet.

There is now a separate option to enable the collapsible sidebar navigation. This means that someone can benefit from the collapsible top level nav items without splitting their content into multiple pages. Registers is an example of a site that would benefit from this.

PR #28, opened 2018-06-26
GTD-1 Search

lewisnyman:
Fixes alphagov/tech-docs-template#66

This integrates middleman-search to generate a Lunr.js index. On the frontend it's a custom Lunr.js implementation.

You'll need a [forked version of middleman-search](alphagov/middleman-search) to run this locally. To do this you'll need to add this line to the Gemfile in your tech docs folders:
`gem 'middleman-search', :git => "git://github.com/alphagov/middleman-search.git"`

Another issue here is that middleman-search generates an index of pages, so this functionality is only useful on multi page sites. We plan to fork middleman-search and build an index of sub headings as well as pages.

In the meantime, I'd like to get this merged as soon as possible behind a config flag, so teams can start to use it and provide feedback.

km3hmen8hw


lib/assets/javascripts/_modules/search.js
```diff
@@ -0,0 +1,171 @@
...
+
+      // Closing the search results, move focus back to the search input
+      $searchResultsClose.on('click', function(e) {
+        e.preventDefault();
+        $searchInput.focus();
+        hideResults();
+      });
+    }
+
+    function searchIndex(query) {
```
MatMoore: I think `searchIndex` is a weird name for this function, since it doesn't return the search index, or do any indexing, it just does a search. Maybe `getResults` would be clearer.

lib/assets/javascripts/_modules/search.js
```diff
@@ -0,0 +1,171 @@
...
+        output += '<p>Nothing found</p>';
+      }
+      output += '<ul>';
+      for(var index in results) {
+        var result = results[index];
+        var content = processContent(result.content, query);
+        output += '<li class="search-result">';
+        output += '<h3 class="search-result__title">';
+        output += '<a href="' + result.url + '">';
+        output += result.title;
```
MatMoore: This could inject HTML into the page. It's likely that at some point it will contain characters that require HTML encoding like &<> which would break the markup of the page. Can we use a templating language instead of string concatenation?

lib/assets/javascripts/_modules/search.js
```diff
@@ -0,0 +1,171 @@
...
+      $searchResults.html( output );
+    }
+
+    function processContent(content, query) {
+      var output;
+      content = '<div>'+  content + '</div>'; 
+      content = $(content).mark(query);
+
+      // Split content by sentence.
+      var sentences = content.html().replace(/(\.+|\:|\!|\?|\r|\n)(\"*|\'*|\)*|}*|]*)/gm, "|").split("|");
```
MatMoore: Can you create tests for this code that generates the snippets?

I think I understand what this is doing, but it's not super clear what the intended behaviour is, and if we end up changing that regex (for example) I wouldn't want to break edge cases that you've already considered.

I expect we may want to truncate these further since matching sentences could be quite long, but I'd rather wait and see if it's actually an issue.

lib/assets/javascripts/_modules/search.js
```diff
@@ -0,0 +1,171 @@
...
+        output = ' … ' + selectedSentences.join(' … ') + ' … ';
+      }
+      return output;        
+    }
+
+    // Default text is to display the number of search results
+    function updateTitle(text) {
+      if(typeof text == "undefined") {
+        var count = results.length;
+        var text = count + ' Results';
```
MatMoore: minor stylistic thing - should "results" be lowercased here?

lib/assets/javascripts/_modules/search.js
lewisnyman: > I expect we may want to truncate these further since matching sentences could be quite long, but I'd rather wait and see if it's actually an issue.

Yeah I can see that being an issue. It's going to be tricky to truncate without accidentally removing required context. Let's see what comes up.

lib/assets/javascripts/_modules/search.js
lewisnyman: I started looking into potential templating systems but when writing the unit test I realised that the index we generate with lunrjs doesn't include any HTML.

I'm not against adding a templating system, but I'd rather not risk delaying this merge on a potential bike shed. How about we open a new issue for that?

spec/javascripts/search-spec.js
```diff
@@ -0,0 +1,22 @@
...
+  var processedContent;
+  var expectedResults;
+
+  beforeEach(function() {
+    module = new GOVUK.Modules.Search();
+    processedContent = module.processContent(content, query);
+  });
+
+  it('returns concatenated sentences that contain the search query', function() {
+    expect(processedContent).toEqual(expectedResults)
```
tijmenb: I'm was confused by this because `expectedResults` is used before it's set. Would it be better to move everything into the `it` block? I'm still not entirely sure we're not accidentally doing `expect(null).toEqual(null)` here.

govuk_tech_docs.gemspec
```diff
@@ -29,9 +29,11 @@ Gem::Specification.new do |spec|
   spec.add_dependency "middleman-livereload"
   spec.add_dependency "middleman-sprockets", "~> 4.0.0"
   spec.add_dependency "middleman-syntax", "~> 3.0.0"
+  spec.add_dependency "middleman-search"
```
tijmenb: Does this conflict with the statement in the `Gemfile`?

docs/configuration.md
```diff
@@ -28,6 +28,14 @@ Adds a [Google Site Verification code](https://support.google.com/webmasters/ans
 google_site_verification: TvDTuyvdstyusadrCSDrctyd
 ```
 
+## `search`
+
+Enables search functionality. This indexes pages only and is not recommended for single-page sites.
+
+```yaml
+search: true
```
tijmenb: Perhaps this can be `enable_search` for consistency with other boolean options like `show_govuk_logo`?

Gemfile
```diff
@@ -2,3 +2,5 @@ source 'https://rubygems.org'
 
 # Specify your gem's dependencies in govuk_tech_docs.gemspec
 gemspec
+
+gem 'middleman-search', :git => "git://github.com/alphagov/middleman-search.git"
```
tijmenb: We could add instructions to add this gem to `docs/configuration.md`. We might keep even want to keep and not require `middleman-search` by default so sites that don't use search don't need to download it.

govuk_tech_docs.gemspec
lewisnyman: It does not. We can't specify a git path in the spec file, but we can in the Gemfile. We don't need to add this line in the spec file but I think it's better to only use the Gemfile to override the gem location. https://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

Gemfile
lewisnyman: Ok, adding this line to their tech docs gemfile would be another step to activating search? It feels like something to avoid. I don't expect tech writers to understand gemfiles and what they do, it adds another step that could break the build if done incorrectly.  It's your call.

lewisnyman:
@MatMoore I've created an issue on our Jira board to add analytics events. It would be good to see the event labels used on the gov.uk search so we can keep them consistent. I do think there is a little complexity around what is classified a 'search term' for a search-as-you-type ui. I'd rather add this separately so we can get this PR merged as soon as possible this week.

I've added some additional tests, let me know if there's anything else I can do to get this merged.

jonathanglassman:
@tijmenb Rafal has reviewed the front end, and yourself and @MatMoore have reviewed the back end. Are you happy to approve and merge?

Gemfile
lewisnyman: 🚨Adding this line is now a requirement to activating search.

PR #31, opened 2018-07-13
Removed Base64 encoded fonts

Nooshu:
Removed the Base64 encoded fonts to remove this data from the browsers critical path. This has been replaced by standard `@font-face` declarations. More information of this change can be seen in [PR 341](https://github.com/alphagov/govuk_template/pull/341) from govuk_template.

I've tested these changes in the example application and they work as expected.


lib/assets/stylesheets/utilities/_fonts.scss
```diff
@@ -1,29 +1,47 @@
-/* Warning: this file is automatically generated by `gds/fonts` - do not edit by hand */
```
tijmenb: Any idea why this is duplicated from https://github.com/alphagov/tech-docs-gem/blob/master/lib/assets/stylesheets/_fonts.scss?

Nooshu: Utilities is the scss file that changes the fonts, the ` /lib/assets/stylesheets/_fonts.scss` looks redundant to me. I will try removing the redundant one and see if anything breaks.

Nooshu: Confirmed that the extra font file is redundant. Removed and squashed the commits.

Nooshu:
Closing this as this is covered by [this PR](#34).

PR #34, opened 2018-07-19
[WIP] Remove GOV.UK frontend toolkit and convert to GOV.UK Frontend

Nooshu:
**See task list of areas left to be integrated**

- [x] Can we use NPM to install Frontend?
- [ ] Check colours all map correctly from Frontend.
- [ ] Move JavaScript from FE into the correct place.
- [x] Use the Phase banner component from Frontend.
- [ ] Use skip link component from Frontend.
- [ ] Adopt new typography classes from Frontend.
- [x] Adopt new spacing scale from Frontend.
- [x] Adopt new header from Frontend.
- [x] Adopt new footer from Frontend.

Related issue: [#154 raised here](https://github.com/alphagov/tech-docs-template/issues/154).

The dependency on [GOV.UK frontend toolkit](https://github.com/alphagov/govuk_frontend_toolkit) has been removed and been replaced by [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend).

Note: this also fixes the [Base64 encoded font issue / PR](https://github.com/alphagov/tech-docs-gem/pull/31) too.

Below are some screenshots comparing the homepage for both on mobile, tablet and desktop:

### Toolkit
mob-toolkit

### Frontend
mob-frontend

### Toolkit
tab-toolkit

### Frontend
tab-frontend

### Toolkit
dt-toolkit

### Frontend
dt-frontend


example/Gemfile
```diff
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 
 gem 'govuk_tech_docs', path: '..'
+gem 'middleman-search', git: 'git://github.com/alphagov/middleman-search.git'
```
36degrees: Don't think this change is related?

lib/assets/stylesheets/_core.scss
```diff
@@ -36,12 +57,12 @@ $desktop-breakpoint: 992px !default;
 @import "accessibility";
 
 html, body {
```
36degrees: Ideally the template should be updated to use the classes from https://github.com/alphagov/govuk-frontend/blob/master/src/core/_template.scss

lib/assets/stylesheets/_core.scss
```diff
@@ -52,17 +73,17 @@ html, body, div, h1, h2, h3, h4, h5, h6, article, aside, footer, header, hgroup,
 }
 
 a:link {
-  color: $link-colour;
+  color: govuk-colour("blue");
```
36degrees: Suggest enabling global link and paragraph styling in Frontend and removing these link rules.

lib/assets/stylesheets/govuk-frontend/README.md
```diff
@@ -0,0 +1,75 @@
+# The `src` directory
```
36degrees: Ideally GOV.UK Frontend would be installed via `npm install` to avoid having to vendor it in.

If that's not possible, I strongly recommend using the `package` folder from the repo rather than the contents of `source`, as the Sass in source is not vendor-prefixed, and the source folder includes e.g. all the tests.

lib/assets/stylesheets/govuk-frontend/all.js
```diff
@@ -0,0 +1,53 @@
+import { nodeListForEach } from './common'
```
36degrees: The javascript from Frontend is inside a folder called 'stylesheets'. Suggest moving the Frontend dependency into a vendor directory or something (assuming installing it via `npm install` is not possible)

lib/assets/stylesheets/modules/_anchored-heading.scss
```diff
@@ -39,13 +39,13 @@
   }
 
   &:hover &__icon {
-    @include media(tablet) {
+    	@include govuk-media-query($from: tablet) {
```
36degrees: Is this change in indentation intentional…?

lib/assets/stylesheets/modules/_anchored-heading.scss
```diff
@@ -39,13 +39,13 @@
...
       opacity: 1;
     }
   }
 
   &__icon:focus {
-    @include media(tablet) {
+    	@include govuk-media-query($from: tablet) {
```
36degrees: Is this change in indentation intentional…?

lib/assets/stylesheets/modules/_collapsible.scss
```diff
@@ -14,7 +14,7 @@
 .collapsible__toggle {
   position: absolute;
   top: 0;
-  right: -25px;
+  right: $govuk-gutter-half * -1;
```
36degrees: The `gutter` variables from Frontend are only for the space in between the grid – ideally this would use the spacing scale, or continue to use a hardcoded value.

(unless this is dependant on the grid, in which case I'd add a comment explaining)

lib/assets/stylesheets/modules/_contribution-banner.scss
```diff
@@ -1,17 +1,17 @@
 .contribution-banner {
   padding: 0;
   max-width: 40em;
-  margin-top: $gutter * 2;
-  margin-left: $gutter-half;
-  margin-right: $gutter-half;
+  margin-top: $govuk-gutter * 2;
```
36degrees: Again, spacing scale.

lib/assets/stylesheets/modules/_footer.scss
```diff
@@ -12,30 +12,30 @@
  *     <a class="footer__licence-logo" href="https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/" rel="license">Open Government Licence</a>
```
36degrees: Any reason we can't use the footer from Frontend?

lib/assets/stylesheets/modules/_header.scss
```diff
@@ -42,24 +42,24 @@
 $active-nav-color: #1d8feb;
 
 .header {
-  @extend %contain-floats;
+  @include govuk-clearfix;
```
36degrees: Any reason we can't use the header from Frontend?

lib/assets/stylesheets/modules/_header.scss
```diff
@@ -167,21 +167,22 @@ $active-nav-color: #1d8feb;
...
     // We want the same font size on mobile as tablet, so we can't use bold-16.
-    @include _core-font-generator(
-      $font-size: 16px,
-      $font-size-640: 16px,
-      $font-size-print: 12pt,
-      $line-height: (20 / 16),
-      $line-height-640: (20 / 16),
-      $font-weight: 700
-    );
+    // @include _core-font-generator(
```
36degrees: Remove?

lib/assets/stylesheets/_core.scss
36degrees: If not, this should be $govuk-link-colour

lib/assets/stylesheets/_core.scss
```diff
@@ -52,17 +73,17 @@ html, body, div, h1, h2, h3, h4, h5, h6, article, aside, footer, header, hgroup,
...
 }
 
 a:visited {
-  color: $link-visited-colour;
+  color: $govuk-link-visited-colour;
 }
 
 a:hover {
-  color: $link-hover-colour;
+  color: govuk-colour("light-blue");
```
36degrees: $govuk-link-hover-colour

lib/assets/stylesheets/modules/_header.scss
```diff
@@ -134,25 +134,25 @@ $active-nav-color: #1d8feb;
...
-        border-bottom: 1px solid $white;
+        border-bottom: 1px solid govuk-colour("white");
       }
     }
   }
 
   @include print {
     a:link, a:visited {
-      color: $black;
+      color: govuk-colour("black");
```
36degrees: $govuk-print-text-colour (which is actually #000!)

lib/assets/stylesheets/modules/_contribution-banner.scss
36degrees: Lots more occurrences of this throughout – not going to comment on all of them.

lib/assets/stylesheets/modules/_phase-banner.scss
```diff
@@ -9,14 +9,14 @@
  */
 
 .phase-banner {
-  @include bold-14;
+  @include govuk-font(14, $weight: bold);
```
36degrees: Can we use the phase banner component from Frontend?

lib/assets/stylesheets/modules/_skip-link.scss
```diff
@@ -19,13 +19,13 @@
   &:focus {
```
36degrees: Can we use the skip link component from Frontend?

lib/assets/stylesheets/modules/_technical-documentation.scss
```diff
@@ -42,65 +46,65 @@
 
   // Headings
   h1 {
```
36degrees: Can we use the new typography classes from Frontend? We do this in the Design System [like this](https://github.com/alphagov/govuk-design-system/blob/master/src/stylesheets/main.scss#L302-L351)

36degrees (commented):
This feels like it's adopting the framework but avoiding adopting any of the conventions - is that intentional?

Nooshu:
> This feels like it's adopting the framework but avoiding adopting any of the conventions - is that intentional?

At the moment yes to get it to a state where it looks and functions fine. I've raised the PR for general feedback and also in the hope that someone will be able to pick up some of the other areas of integration as it is quite a time consuming task.

lib/assets/stylesheets/modules/_collapsible.scss
lewisnyman: This was the only other visual regression I found, this changes the value and moves the placement of the toggle button. I'd recommend using the previous value.

lewisnyman (changes requested):
It's great to see this being implemented., thanks for working on this.
I ran a quick visual regression test. It looks like the new logo implementation increases the logo height by one pixel, I don't know if that's worth worrying about or not.

Nooshu:
Currently in the process or integrating the DS components into the tech docs gem. Integration of the header component / phase banner has brought up a couple of design considerations. 

Is the plan to keep the header / navigation the same but directly bring code across from the DS? If so the output will look like this:
screen shot 2018-08-09 at 14 38 53

Or should we be considering using the navigation that is now being used for the Design System:
screen shot 2018-08-09 at 14 39 07

I personally think the overall design on the DS nav / header looks better than "Header with service name and navigation", but I'm happy to go with whatever the community would prefer.

MatMoore:
@Nooshu what's the name of that component the design system uses? 

I like how it works on the design system site, but I'm wondering how it would work for sites where the tech docs is only part of the service. For example on registers [this page](https://www.registers.service.gov.uk/registers) has a different layout to [the documentation part](https://docs.registers.service.gov.uk/#gov-uk-registers-quick-start-guide), which is a bit jarring, but the top level navigation still looks the same wherever you are.

Also on the [GDS way](https://gds-way.cloudapps.digital/#the-gds-way) site, one of the top level links is an external link, which might not work well with the new thing.

Nooshu:
@MatMoore It is the "Header with service name and navigation" which can be seen on [this page](https://design-system.service.gov.uk/components/header/). I did wonder if it may not work for all documentation, so it is probably a safer to stick with what we know.

Nooshu:
Pushing latest updates:

- NPM now being used to pull in Frontend. This may be temporary until a discussion around the Ruby pipeline is had.
- DS footer now being user.
- Initial version of the DS Header is being used. This will be swapped out with a version that closely matches the current setup (after discussion with @36degrees and @dashouse).
- DS Phase banner has been integrated, this may be removed depending on what happens with the updated header.
- DS JS has been copied across to the assets folder as a temporary measure until the Ruby pipeline integration has been discussed.

**Issues**:
- [ ] Small issue to be solved with horizontal padding when moving between breakpoints with the new footer.
- [ ] Two sets of JS detection CSS classes currently in place. Modernizr and DS systems. Modernizr is used extensively for the overall layout, this could be used in the DS header with a little tweaking (`js-enabled` to 'js').

bendyson:
@Nooshu Just out of interest, is this still being worked on?

Nooshu:
@bendyson not at the moment I'm afraid. Leaving the information here and I'm hoping someone else will have the capacity to finish it.

lib/assets/stylesheets/_core.scss
tlwr: I'm going to raise all of your suggestions as GH suggestions

lib/assets/stylesheets/_core.scss
```diff
@@ -52,17 +72,17 @@ html, body, div, h1, h2, h3, h4, h5, h6, article, aside, footer, header, hgroup,
 }
 
 a:link {
-  color: $link-colour;
+  color: govuk-colour("blue");
```
tlwr:
```suggestion
  color: $govuk-link-colour
```

lib/assets/stylesheets/_core.scss
```diff
@@ -52,17 +72,17 @@ html, body, div, h1, h2, h3, h4, h5, h6, article, aside, footer, header, hgroup,
...
 }
 
 a:visited {
-  color: $link-visited-colour;
+  color: $govuk-link-visited-colour;
 }
 
 a:hover {
-  color: $link-hover-colour;
+  color: govuk-colour("light-blue");
```
tlwr:
```suggestion
  color: $govuk-link-hover-colour;
```

lib/assets/stylesheets/_core.scss
```diff
@@ -52,17 +72,17 @@ html, body, div, h1, h2, h3, h4, h5, h6, article, aside, footer, header, hgroup,
...
 }
 
 a:active {
-  color: $link-active-colour;
+  color: govuk-colour("light-blue");
```
tlwr:
```suggestion
  color: $govuk-link-active-colour;
```

tijmenb:
This is superseded by https://github.com/alphagov/tech-docs-gem/pull/89 now, thanks @Nooshu!

PR #35, opened 2018-07-19
Use fork of `middleman-search`

samcrang:
I'm not entirely sure of the reasons why the fork has to be used but according to @lewisnyman this is required for now.


lewisnyman:
Ah good catch. I never added this line to the example inside the gem, so anyone using it for development will get this error.

It only affects `middleman serve`. See the reason here: https://github.com/alphagov/tech-docs-template/issues/66#issuecomment-347198601

PR #39, opened 2018-07-30
Search Google Analytics tracking

lewisnyman:
This PR adds event tracking on search result clicks and on search terms, there's a wait time of a second after a user has finished typing before an event is sent. This might be too low and we might need to increase this once we have real usage.

### DIFF
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -67,6 +67,20 @@
         $searchInput.focus();
         hideResults();
       });
+
+      // Attach analytics events to search result clicks
+      if(window.ga) {
+        $searchResults.on('click', '.search-result__title a', function() {
+          var href = $(this).attr('href');
+          ga('send', {
+            hitType: 'event',
+            eventCategory: 'Search result',
+            eventAction: 'click',
+            eventLabel: href,
+            transport: 'beacon'
+          });
+        });
+      }
     }
 
     function getResults(query) {
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -15,6 +15,8 @@
     var $searchResultsWrapper;
     var $searchResultsClose;
     var results;
+    var query;
+    var queryTimer;
     var maxSearchEntries = 10;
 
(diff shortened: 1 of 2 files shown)


lib/assets/javascripts/_modules/search.js
```diff
@@ -50,12 +52,16 @@
       // Search functionality on search text input
       $searchInput.on('input', function (e) {
         e.preventDefault();
-        var query = $(this).val();
+        query = $(this).val();
```
MatMoore: Do we need to do any sanitisation on this before sending it to GA? We do scrub things that look like PII from data sent to GA in other places, eg https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/analytics/analytics.js

I'm not aware of any uses where I'd expect users to type in PII but as this is a template it's hard to foresee what it will be used for.

I would suggest either making this functionality opt-in, or adding some explanation of what it does to the [ga-tracking-id](https://alphagov.github.io/tech-docs-manual/#ga-tracking-id) documentation, so that site creators are aware of this.

lib/assets/javascripts/_modules/search.js
```diff
@@ -72,6 +78,20 @@
         $searchInput.focus();
         hideResults();
       });
+
+      // Attach analytics events to search result clicks
+      if(window.ga) {
+        $searchResults.on('click', '.search-result__title a', function() {
+          var href = $(this).attr('href');
+          ga('send', {
```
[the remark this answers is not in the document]
lewisnyman: Yeah this would be useful. I think your best bet here is using the [event flow report](https://support.google.com/analytics/answer/2521316?hl=en). With a custom "users with search event" segment, you should be able to get a useful view.

lib/assets/javascripts/_modules/search.js
lewisnyman: That's a good idea. I've borrowed that stripPII function. I'll leave the full integration for #34.

I'll talk to @jonathanglassman tomorrow about making sure all the new GA tracking events are documented. 

I've also opened an issue about the date regex: alphagov/govuk_frontend_toolkit#469

PR #41, opened 2018-07-31
Remove stopword filter from Lunr index generation in middleman-search.

lewisnyman:
After some testing with registers I found that alphagov/tech-docs-gem#38 was only half a fix.
alphagov/registers-tech-docs#109. As the lunr index is pre-generated by middleman-search, we also need to remove the stopwordfilter there to return useful results for `get`.

I've added this config option in the alphagov fork. See: alphagov/middleman-search#1

### DIFF
--- a/lib/govuk_tech_docs.rb
+++ b/lib/govuk_tech_docs.rb
@@ -102,6 +102,10 @@ module GovukTechDocs
           content: { boost: 50, store: true },
           url:     { index: false, store: true },
         }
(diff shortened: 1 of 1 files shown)


MatMoore:
How come in https://github.com/alphagov/tech-docs-gem/pull/38 you replace the stopword filter, but here it's removed entirely? Shouldn't it be the same when indexing and searching?

PR #51, opened 2018-09-12
Prepare 1.6.0 release

MatMoore:
- [Changelog](https://github.com/alphagov/tech-docs-gem/blob/c4356aaba5710c5526f27a328509210ac062ec4e/CHANGELOG.md)
- [Diffs since 1.5.0](https://github.com/alphagov/tech-docs-gem/compare/5f8f7e7d91c45e26d7c7a11752e408002461b871...master)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,55 @@
 
 ## Unreleased
 
+## 1.6.0
+
+Version 1.6.0 adds API reference generation, and improves the search function.
+
+### New feature: API reference
+
+Specify an OpenAPI spec file using `api_path`, for example: `api_path: source/pets.yml`.
+
+To output the generated content in the page of your choice, write `api>` in any markdown parsed file (.md, .html.md, .html.md.erb). You can also specify to just print a specific path: `api> /pets`
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,55 @@
 
 ## Unreleased
 
+## 1.5.1
```
tijmenb: Should this be 1.6.0 if there's a new feature?

MatMoore: Are we following https://semver.org/ ? If so then I think it's a minor change since everything should be backwards compatible.

```
Given a version number MAJOR.MINOR.PATCH, increment the:

MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards-compatible manner, and
PATCH version when you make backwards-compatible bug fixes.
Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.
```

MatMoore: Oh sorry, just read the actual number and you're right. This should be 1.6.0.

MatMoore:
Have just changed this to 1.6.0 so will merge when the tests pass.

PR #54, opened 2018-10-24
Fix hanging middleman

vixus0:
There is currently an issue with middleman-autoprefixer which causes
middleman to hang: middleman/middleman-autoprefixer#33

When this happens, middleman no longer responds to requests and does not
refresh on source changes.

Moving the `activate :autoprefixer` command into the `context.configure :build` block seems to fix this and middleman behaves normally.


bahmady:
Is it possible to get a new incremental release out with this fix? I know a lot of people struggle with this and it's great to finally have a fix.

tijmenb:
This has now been released!

PR #60, opened 2018-11-30
Customise "Report issue" URL

vixus0:
By default, clicking "Report issue" in the contribution banner at the
bottom of the page (if turned on) will link the user to creating an
issue in the docs' GitHub repo. Some end-users of the tech-docs gem may
instead want to link to a page containing a form or allow the user to send an email.

End-users can achieve this by adding

```yaml
source_urls:
  report_issue_url: https://some-other-site/support-form
```

to their `config/tech-docs.yml` file.

I've added tests for this behaviour in
`spec/govuk_tech_docs/source_urls_spec.rb`.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -139,6 +139,18 @@ show_contribution_banner: true
 github_repo: alphagov/example-repo
 ```
 
+## `source_urls`
+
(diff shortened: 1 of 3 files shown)


bravegrape:
@bahmady @jonathanglassman

PR #62, opened 2019-01-03
Fix sidebar highlighting for single page navigation

klssmith:
Items in the left navigation bar were no longer being highlighted when
active for single page sites (this still worked for multipage sites.)
This is because the `href` attribute for the table of content items has
a different format for single page sites and multipage sites. We were
only looking at the format for multipage table of content items when
deciding which item to highlight, so this commit now ensures we look at
both formats.


jonathanglassman:
@tijmenb any chance you could review?

PR #63, opened 2019-01-08
Remove sidebar pane if there is no sidebar

selfthinker:
When using the `core` layout, the standard sidebar isn't used and you can put anything in it. But if you don't use the sidebar (with `content_for :sidebar`) at all, the empty pane still shows and the content is still further to the right.

tech-docs-no-sidebar-before

This fixes that by removing the sidebar pane if there is no sidebar.

tech-docs-no-sidebar-after

There is a slight chance that this is not 100% backwards compatible in case someone wanted no sidebar content but still wanted the pane to show. I assume that's extremely unlikely, though.


selfthinker:
One downside is that the issue of alphagov/tech-docs-template#123 becomes more apparent with this change as you cannot click the "Menu" toggle in mobile view anymore when the "Table of contents" is missing to add the required space.

tech-docs-cannot-click-menu

selfthinker:
I have fixed that unclickable Menu toggle (although it only partly fixes the other linked issue).

selfthinker:
Adding an example page makes sense, I can do that.

I initially had an alternative way of doing this via a frontmatter variable. I added something to the documentation in that case. But when I did this version instead, I removed the documentation again, because this is more of a bug fix than a new feature. At least I assume that most people would expect a sidebar to not be there if they didn't add any sidebar. That's why I think documenting this behaviour would be confusing.

selfthinker:
Actually, because the example in the documentation only has one example for when you put something else into the sidebar, adding something with a non-existing sidebar **does** make sense. I'll do that.

selfthinker:
Done. I've added the example page, updated the documentation and added a test (not just for the lack of a sidebar but also to test if there is a sidebar on the homepage).

PR #64, opened 2019-01-08
Add possibility to add something to the head

selfthinker:
This makes it possible to inject anything, like further styles, scripts or meta data, into the `<head>` with `content_for :head`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,3 +3,6 @@ rvm:
   - 2.4.2
 script:
   - bundle exec rake
+before_install:
+  - rvm @global do gem uninstall bundler -a -x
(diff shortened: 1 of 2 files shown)


jonathanglassman:
@selfthinker is there a particular user need you have in mind for this change?

selfthinker:
Yes, I plan to move everything related to accessibility personas into one repo. That will include bad and good example pages. I would need to "break" something only on particular pages like we do on https://empathy-lab-personas.herokuapp.com/. (I will send you the password to that separately.)

jonathanglassman:
Ok, that makes complete sense, thanks for the clarification. I've asked Tijmen for a review in case he is concerned about anything from a technical point of view.

PR #65, opened 2019-01-08
Don't be picky about bundler versions

tijmenb:
By requiring bundler 1 we confuse Travis because by default, rvm now installs Bundler version 2. This causes all PRs to fail on CI (https://github.com/alphagov/tech-docs-gem/pull/64 https://github.com/alphagov/tech-docs-gem/pull/63).


tijmenb:
This doesn't work because of subdependencies.

PR #67, opened 2019-01-17
Automate releases to Rubygems

tijmenb:
This sets up a deployment on Travis, which hopefully means it'll release a new version every time we merge a new version to master. This prevents the project from being dependent on me to release new versions.

We've encrypted the Rubygems API key. The API key belongs to the https://rubygems.org/profiles/govuk user. Note that this user belongs to the GOV.UK programme, which has no formal relationship to the tech docs template. However until there's a team officially owning the project, we'll have to do it like this.

Thanks to @36degrees for helping me understand Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,11 @@ script:
 before_install:
   - rvm @global do gem uninstall bundler -a -x
   - rvm @global do gem install bundler -v 1.17.3
+deploy:
+  provider: rubygems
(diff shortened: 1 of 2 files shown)


jonathanglassman:
Thanks @tijmenb. Would someone still need to add in changes to the change log and increment the version in the `version.rb` file?

tijmenb:
@jonathanglassman yes, that should be done in a separate PR.

PR #69, opened 2019-01-18
Make specific folder to be excempt from minification

selfthinker:
This makes it possible to serve and build CSS and JS files in their raw, unminified and unmodified form.
Every CSS and JS file in the `raw_assets` folder will be ignored by the minifiers.

My use case:
I need to add downloadable and installable [userscripts](https://en.wikipedia.org/wiki/Userscript) and [userstyles](https://en.wikipedia.org/wiki/Stylish) to my project (alphagov/accessibility-personas). But the minification strips the comments which are essential for them to work.

I have not documented this yet because it doesn't really fit anywhere. (It's neither a configuration option, nor a frontmatter setting, nor has it anything to do with page expiry.) Shall I just create a new Markdown file?

### DIFF
--- a/lib/govuk_tech_docs.rb
+++ b/lib/govuk_tech_docs.rb
@@ -52,8 +52,8 @@ module GovukTechDocs
 
     context.configure :build do
       activate :autoprefixer
-      activate :minify_css
-      activate :minify_javascript
+      activate :minify_css, ignore: ['/raw_assets/*']
+      activate :minify_javascript, ignore: ['/raw_assets/*']
     end
 
     context.config[:tech_docs] = YAML.load_file('config/tech-docs.yml').with_indifferent_access


selfthinker:
I've just played around with this a bit.

I found that multiline comments would not work for userstyles as Stylus (the Chrome extension interpreting the userstyles) cannot find the essential metadata. (I haven't tested userscripts, but I assume the result would be the same.)

But I also found that multiline comments still get minified, it's the `!` in the font comment which makes it not being minified, e.g. `/*! ==UserStyle==` works.

But it is good practice to not minify any userscripts and userstyles as you're basically asking strangers to trust your code and install it in their browser. It's better if they can easily read and double check the whole code before installing. That's why I'd prefer to let minification ignore them all.

selfthinker:
Actually, another solution would be to point to the files on GitHub, e.g. https://github.com/alphagov/accessibility-personas/raw/master/source/javascripts/wobbly-mouse.user.js
That definitely works as well.

tijmenb:
👍Thanks for the explanation! Feel free to merge.

PR #71, opened 2019-02-06
Add github_branch config option

timja:
We're publishing to github pages from an organisation site (hmcts.github.io)
Organisation sites require that your master branch publishes the assets

So we've created another branch `source` which has our source files

I would rather not set `source_url` on every page, it'll end up being wrong in some places

Would you be happy to take this configuration option?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 1.7.1
+
+### Make `github_branch` configurable
+
+Previously it was hard-coded to master unless you set the source_url on every page
+
 ## 1.7.0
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 1.7.1
```
tijmenb:
```suggestion
## Unreleased
```

This allows us to bundle multiple changes into a release (which could be `1.8.0` or even `2.0.0`).

timja:
@tijmenb Done and conflict in changelog resolved

timja:
Feel free to amend changelog as you wish

PR #78, opened 2019-03-19
Add warning text extension

vixus0:
Now you too can warn people as recommended by the GOV.UK Design System:
https://design-system.service.gov.uk/components/warning-text/

All you have to do is include: `<%= warning_text('Look out!') %>` in
your ERB content.

### DIFF
--- a/lib/assets/stylesheets/modules/_warning-text.scss
+++ b/lib/assets/stylesheets/modules/_warning-text.scss
@@ -1,0 +1,73 @@
+.govuk-warning-text {
+  font-family: nta,Arial,sans-serif;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+  font-weight: 400;
+  font-size: 16px;
+  font-size: 1rem;
+  line-height: 1.25;
+  color: #0b0c0c;
+  position: relative;
+  margin-bottom: 20px;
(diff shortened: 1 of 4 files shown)


bravegrape:
[Get warning box from Design System](https://trello.com/c/tdUdwRcg/206-get-warning-box-from-design-system)

bravegrape:
When this feature goes live, we should also add documentation to https://tdt-documentation.london.cloudapps.digital/ 👌

lib/assets/stylesheets/modules/_warning-text.scss
```diff
@@ -0,0 +1,73 @@
+.govuk-warning-text {
+  font-family: nta,Arial,sans-serif;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+  font-weight: 400;
+  font-size: 16px;
+  font-size: 1rem;
+  line-height: 1.25;
+  color: #0b0c0c;
```
jakubmiarka: we should probably be using colours from the [palette](https://github.com/alphagov/tech-docs-gem/blob/b650d313b893af4990603c1cff917ab1776a0d65/lib/assets/stylesheets/govuk_frontend_toolkit/colours/_palette.scss) and other stuff.

PR #83, opened 2019-04-05
Move content across from alphagov/tdt-documentation

### LINKED ISSUE #491: [Enhancement]: Documentation should be published, and it should belong in the gem
### What should change?

This issue links up PR #83 

### User need

There needs to be documentation that users can access to understand how to use the `tech-docs-gem` and the `tech-docs-template`.  Since the original documentation site was closed down it is not publicly available.

It has currently been published via GitHub Pages on the existing repo, but would be better moved to this gem to simplify maintenance and make it easier to find

### Proposed solution

Move the content to here, and then publish from here

bravegrape:
### How to review

**TO DO**
- [x] move content from `alphagov/tdt-documentation`
- [x] make sure the content has the latest updates from `docs`
- [x] put all images in a separate `docs/source/images` folder
- [x] make sure images are referenced correctly from the new location
- [ ] remove fingers
- [ ] write a proper PR description
- [ ] sort out `.gitignore`
- [ ] fix links referring to old `docs` folder, for example in `CONTRIBUTE.md`
- [ ] check that all links still work
- [ ] deploy to PaaS
- [ ] set up Travis to deploy automatically
- [ ] check all content again before archiving, in case other changes have been made


huwd:
This needs another review and an update I think,  I'll see if I can find time.
bringing the docs over and removing broken / dead TDT links sounds good - but I think we need to update the actual guidance that comes across, I'm seeing some PaaS references and lots of ruby v2.x stuff,

I'll go through line by line but I think this is a minor update after v4

JC-DSIT:
closing this in favour of https://github.com/alphagov/tech-docs-gem/issues/491

PR #86, opened 2019-06-10
Keep search results when opening in new tabs

ChrisBAshton:
| Before | After |
|-------|-------|
|new tab behaviour hides search result|new tab behaviour keeps search results in place|

Fixes an issue whereby if you 'CMD + Click' on search results, the search results are closed.

(This made it difficult to line up several useful looking results in new tabs; you'd either have to click
on the search input again and re-trigger the search to re-show the results and
enable opening a result in a new tab, or give up on tabs altogether and navigate
in your current window).
    
The selector '.toc' contains the search input, the sidebar and the search results
themselves. So calling 'hideResults' when '.toc a' is clicked meant that the
search results were hidden the moment you click on a search result.
    
The selector '#toc' contains only the search input and sidebar - NOT the search
results, so we can safely call 'hideResults' when '#toc a' is clicked. This means
we can right-click on search results and open them in new tabs, without the results
being cleared. It also means we haven't broken the existing behaviour (which is that
if a navigation link in the sidebar is clicked, the search results should be hidden - as
the search results are overlayed on top).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,10 @@ removes the need for projects to point to the alphagov github repository in
 their Gemfiles. We will continue to maintain our own fork until the changes
 are merged into the upstream project.
 
+## Unreleased
+
+Fixes bug where search results disappear when opening results in a new tab, making it difficult to open several results in a batch (PR #86).
+
 ## 1.6.2
 
 This version fixes an issue with Middleman hanging (PR #54). It also allows the API reference page to include multiple servers and their descriptions (PR #53).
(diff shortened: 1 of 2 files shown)


benthorner:
I'm wondering (somewhat lazily) what the behaviour is when you simply click on a search result now? I guess it'll try to render the page, but will the search results still be shown on top?

ChrisBAshton:
It works like a normal link, i.e. loads a new page (so the whole page is re-rendered, without search results).

ChrisBAshton:
Thanks! Although I can't merge because "Only reviews by reviewers with write access count toward mergeability" (I also wasn't able to push up a branch, had to do a fork).

I wonder if the permissions are borked since the GitHub team deletion last week? 🤔

tijmenb:
👍 I've added this repo to the GOV.UK team, so feel free to merge.

ChrisBAshton:
Done 👍

PR #91, opened 2019-07-18
Add StandardJS

### LINKED ISSUE #90: Add Standard.js to lint all the JavaScript
In the [GDS Way](https://gds-way.cloudapps.digital/manuals/programming-languages/js.html#linting), it mentions using Standard to lint our projects yet this one does not use.

aliuk2012:
Fixes #90 

- Adds Standard as a dev dependency and adds it to the pipeline
- Fixes general styling issues identified by Standard
- Removes unused variables identified by Standard


package.json
```diff
@@ -7,5 +7,25 @@
   },
   "devDependencies": {
     "standard": "^13.0.2"
+  },
+  "standard": {
+    "globals": [
```
ChrisBAshton: Is it worth saying where this list came from (in the commit message)?

lib/assets/javascripts/_modules/collapsible-navigation.js
```diff
@@ -2,18 +2,11 @@
   'use strict'
 
   Modules.CollapsibleNavigation = function () {
-    var $headings;
-    var $listings;
-
-    var $openLink;
-    var $closeLink;
```
ChrisBAshton: This is why linting is awesome! 🎉 Love this.

package.json
aliuk2012: 🤔 I'm not sure if it's worth it.

Most of these are to ignore Jasmine functions (i.e describe, it, assert etc) the others were really just globals refer to in various Javascript files that failed linting because they were not declared in the scripts themselves.

Issue #93, opened 2019-07-31
Use GOV.UK Frontend instead of GOV.UK Frontend Toolkit


tijmenb:
This issue is to keep track of all todos:

- [ ] focus state

PR #94, opened 2019-07-31
Use focus styles from GOV.UK Frontend

tijmenb:
This updates the CSS to use the focus styles from GOV.UK Frontend.

I've used https://design-system.service.gov.uk/get-started/focus-states as a guide.

I've disabled the `$focus-colour` variable in our vendored frontend toolkit to make sure we aren't using it anymore.

Re: https://github.com/alphagov/tech-docs-gem/issues/93

### DIFF
--- a/lib/assets/stylesheets/_accessibility.scss
+++ b/lib/assets/stylesheets/_accessibility.scss
@@ -4,6 +4,5 @@ a {
 }
 
 a:focus {
-  background-color: $focus-colour;
-  outline: 3px solid $focus-colour;
+  @include govuk-focused-text;
 }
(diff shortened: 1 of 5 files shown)


lib/assets/stylesheets/_accessibility.scss
```diff
@@ -4,6 +4,5 @@ a {
 }
 
 a:focus {
-  background-color: $focus-colour;
-  outline: 3px solid $focus-colour;
+  @include govuk-focused-text;
```
tijmenb: @36degrees:

> I’m wary of adding the govuk-focused-text mixin to all links – I think you’re going to end up fighting it in the navigation styles. My recommendation would be to scope that just to links within the markdown-generated content and manually add focus styles to the other components that make up the page.
>
> A lot of the navigation components are going to need updating so that they behave in a similar way to the navigation elsewhere – larger click areas, but with the focus state still tight around the text

tijmenb:
Closing this because of the added complexity of https://github.com/alphagov/tech-docs-gem/pull/94#discussion_r309203760. I'll leave this work to a frontend dev to complete!

PR #95, opened 2019-08-28
Simplify how JavaScript is initialised and slightly improve performance

aliuk2012:
This is part of a bigger piece of work to remove the dependency on jQuery.

There is no need to rely on jQuery's document.ready() method.
By placing application.js at the bottom of the page, the script will be
executed after all preceding DOM is loaded anyway.

There is also a slight performance improvement (HTML parsing and loading
of CSS and images isn't blocked by the scripts)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,13 @@
 
 ## Unreleased
 
+### Fixes
+
+- [Pull request #95: Simplify how JavaScript is initialised and slightly improve performance](https://github.com/alphagov/tech-docs-gem/pull/95)
+
 ## 2.0.10
+
+### Fixes
 
 - [Pull request #160: Make sure IDs on collapsible navigation are unique](https://github.com/alphagov/tech-docs-gem/pull/160)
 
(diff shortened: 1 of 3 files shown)


jonheslop:
Nice! I have been using 

```
  document.addEventListener('DOMContentLoaded', function() {
	// stuff
  });
```

as a vanilla alternative but if you’re sure everything loads in the right order then 👌

lib/source/layouts/core.erb
```diff
@@ -18,6 +18,11 @@
       <meta name="google-site-verification" content="<%= config[:tech_docs][:google_site_verification] %>" />
     <% end %>
 
+<<<<<<< HEAD
```
nickcolley: dang I missed this :(

nickcolley:
I've bumped this ready to merge, I think some newer changes (probably move to the Design System) moved the script location anyway so this feels good to merge.

nickcolley:
Hmm this approach no longer seems to work anymore, going to close for now, will be revisited when/if we remove jQuery. Thanks for the help anyway.

PR #105, opened 2019-09-05
Bundle remaining assets type from GOV.UK Frontend

freesteph:
The Crown logo appears to be missing after upgrading to 2.0.1. Since this bug happens in the gem-cutting process (on Travis) I can't guarantee the fix but am pretty sure it should check out.


jonheslop (approved):
Good spot!

PR #113, opened 2019-09-18
Fix bug in API docs that shows required attributes as optional

tijmenb:
:wave: from [DfE](https://github.com/DFE-Digital)!

This fixes a bug in the [OpenAPI reference renderer](https://tdt-documentation.london.cloudapps.digital/add_an_openapi_specification.html) that affects which attributes of a [schema](https://swagger.io/docs/specification/data-models/) are considered to be required.

For the uninitiated: in OpenAPI, schemas allow you to define what objects/resources/models your API will return. It uses [JSON Schema](https://json-schema.org/) for this.

For example:

```yml
schemas:
  Pet:
    required:
      - id
      - name
    properties:
      id:
        type: integer
        format: int64
      name:
        type: string
      tag:
        type: string
```

This means that any `Pet` objects will have at least an `id` and a `name`, and an optional `tag`.

Currently the logic to display which attributes are required is wrong, because the code expects something like:

```yml
id:
  type: integer
  format: int64
  required: true
```

This isn't valid JSON Schema as far as I can see: 

https://json-schema.org/understanding-json-schema/reference/object.html#required-properties

This commit fixes the bug by making the check look at the `required` array in the schema, rather than the property.

## How to review

[Read the main commit](https://github.com/alphagov/tech-docs-gem/pull/113/commits/5f2fd1429f168070d164b9ee05b261107a02bb50), the other one is a small refactoring to make things clearer.

## Before

Screenshot 2019-09-17 at 22 08 34

## After

Screenshot 2019-09-17 at 22 07 48

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changelog
 
+Fix bug in API docs that shows required attributes as optional  (#113)
+
 ## 2.0.3
 
 Fixes a couple of styling issues: [#100](https://github.com/alphagov/tech-docs-gem/issues/100) and [#106](https://github.com/alphagov/tech-docs-gem/issues/106).
--- a/lib/govuk_tech_docs/api_reference/templates/schema.html.erb
+++ b/lib/govuk_tech_docs/api_reference/templates/schema.html.erb
@@ -1,7 +1,7 @@
 <h3 id="<%= id = 'schema-' + title; id.parameterize %>"><%= title %></h3>
 <%= markdown(schema.description) %>
 <% if properties.any? %>
-<table>
+<table class='<%= id.parameterize %>'>
 <thead>
 <tr><th>Name</th><th>Type</th><th>Required</th><th>Description</th><th>Schema</th></tr>
 </thead>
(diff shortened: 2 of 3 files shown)


tijmenb:
Thanks @36degrees!

I'm actually going to close this, because it was based on a misunderstanding of OpenAPI and JSON Schema. It turns out that OpenAPI "forked" JSON Schema and [made some subtle changes](https://phil.tech/api/2018/03/30/openapi-and-json-schema-divergence/). This is just wonderful.

Thus, my assertion that "required: true" isn't JSON Schema is true, but it's not relevant to OpenAPI. And this fix would actually break support for OpenAPI.

Because we found a number of additional issues with the API reference feature, we've mostly [rewritten it for our tech docs](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training-tech-docs/tree/master/lib/govuk_tech_docs/api_reference). Once we've tested this version with users we'll try to push it back into the tech-docs-gem. 

I haven't been able to find another user of the API reference feature in the tech docs by the way, if you are using it let me know, and we can talk about improving it.

PR #115, opened 2019-09-19
allow configuring footer_links

chrisfarms:
## What

Allow configuring links that display in the footer of the core template form the config:

```
footer_links:
  Page Name: /page/link/url
```

To make this more practical, we also add support for skipping rendering a page in the page tree by setting
`hide_in_navigation: true` frontmatter.

## Why

Some pages (such as accessibility info, privacy info, etc) do not really fit in the main navigation, and we want a way to add this kind of information.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -55,6 +55,17 @@ header_links:
   Documentation: /
 ```
 
+## `footer_links`
+
+Links to show in footer.
+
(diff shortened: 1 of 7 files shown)


bravegrape:
yes, this is very useful - we'll have a nice way to add in the accessibility statements 👌

PR #118, opened 2019-10-04
Tidy up and optimise imports

36degrees:
- Remove unused font CSS files
- Remove unused breakpoints
- Replace `printable` mixin with `govuk-media-query`
- Only include the bits of GOV.UK Frontend we actually need

Reduces the CSS size from 145KB to ~98KB (before gzip)

(I attempted to do the same for the JavaScript, but couldn't get Frontend to play nicely with Sprockets, so that's a job for the future)


lib/assets/stylesheets/_govuk_tech_docs.scss
```diff
@@ -1,6 +1,23 @@
...
+@import "govuk/core/all";
+@import "govuk/objects/all";
+
+@import "govuk/components/footer/footer";
+@import "govuk/components/header/header";
+@import "govuk/components/inset-text/inset-text";
+@import "govuk/components/input/input";
+@import "govuk/components/tag/tag";
+@import "govuk/components/skip-link/skip-link";
+@import "govuk/components/warning-text/warning-text";
```
nickcolley: When I've done this sort of thing in the past I've left in the full component list as comments for other developers ease of use but probably is an opinionated idea.

36degrees: I'm going to leave as-is, mainly because I can see a list like that going out of date relatively quickly as we introduce new components etc.

nickcolley: Yeah that's a good point :+1:

PR #121, opened 2019-10-08
Pin openapi3_parser

36degrees:
The 0.6.0 release is causing builds to fail. Pinning it to 0.5.2 appears to solve this.

The maintainer of openapi3_parser has been made aware, and hopes to have a new version out soon.

### DIFF
--- a/govuk_tech_docs.gemspec
+++ b/govuk_tech_docs.gemspec
@@ -45,7 +45,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "middleman-search-gds"
   spec.add_dependency "nokogiri"
   spec.add_dependency "redcarpet", "~> 3.3.2"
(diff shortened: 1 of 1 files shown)


36degrees:
Mostly raising this to check that it does fix the failing tests currently on master.

If a new release of openapi3_parser is released _soon_, we may not need to merge this.

kevindew:
@36degrees I fixed the underlying bug in openapi_parser with 0.6.1 however I discovered (much to my surprise) that this gem also relies on a method that was removed, so I'd suggest you merge this in.

I'm planning to update the code in this gem for compatibility without this method and will open a fresh PR for that (after thoroughly rasping myself for not realising the significance of a BC break) and can update the gem dependency then too.

Apologies for the inconveniences.

PR #125, opened 2019-10-14
Update openapi_parser

kevindew:
Following up from https://github.com/alphagov/tech-docs-gem/pull/121 this updates openapi_parser to allow the latest version of gem. Since the gem is below version 1 it seemed safe to maintain the pin.

I've also done some cleaning up of some of the Ruby code in `GovukTechDocs::ApiReference::Renderer` to make it more shorter and more consistent with modern Ruby. There's a few dubious things in there that I left as I didn't want to risk creating fresh bugs from "fixing" stuff.

I've also rebased https://github.com/alphagov/tech-docs-gem/pull/113 onto this as that fix would conflict with the changes here.

Edit:

Since opening this @tijmenb closed #113 so the link to that is no longer accurate. I didn't want to re-instate a bug by undoing that so wanted to keep that change here and had already amended his change to use a new method in openapi_parser to check required status.

To document the issue, an [OpenAPI document schema object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#schemaObject) has a required property which [must be an array of strings](https://tools.ietf.org/html/draft-wright-json-schema-validation-00#section-5.15) which is used to indicate what properties of that object are required.

Previously the schema template just checked whether that array was set to determine if a parameter was required: https://github.com/alphagov/tech-docs-gem/blob/a2ba6143342ea9b58fef879a3946500fba71db23/lib/govuk_tech_docs/api_reference/templates/schema.html.erb#L13
This is incorrect as this only tells you whether that parameter requires properties of its own.

Instead we use the [`#requires?`](https://www.rubydoc.info/github/kevindew/openapi3_parser/Openapi3Parser/Node/Schema#requires%3F-instance_method) method on the schema to determine if this property is required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- [Pull request: #125: Fix API docs showing required properties as optional](https://github.com/alphagov/tech-docs-gem/pull/125)
+
 ## 2.2.0
 
 ### Accessibility Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,7 +100,6 @@ You can look at the [2.0.6 milestone](https://github.com/alphagov/tech-docs-gem/
 
 ## 2.0.5
 
-
 Adds [new global configuration option](https://github.com/alphagov/tech-docs-gem/pull/122) that controls whether review banners appear or not at the bottom of pages when expiry dates are set in the frontmatter.
 
 [Fixes the hard-coded service link](https://github.com/alphagov/tech-docs-gem/pull/119) in the header.
--- a/lib/govuk_tech_docs/api_reference/api_reference_renderer.rb
+++ b/lib/govuk_tech_docs/api_reference/api_reference_renderer.rb
@@ -107,12 +107,6 @@ module GovukTechDocs
         @template_responses.result(binding)
       end
 
-      def markdown(text)
-        if text
-          Tilt["markdown"].new(context: @app) { text }.render
-        end
-      end
-
       def json_output(schema)
         properties = schema_properties(schema)
         JSON.pretty_generate(properties)
--- a/lib/govuk_tech_docs/api_reference/api_reference_renderer.rb
+++ b/lib/govuk_tech_docs/api_reference/api_reference_renderer.rb
@@ -18,20 +18,15 @@ module GovukTechDocs
       end
 
       def api_full(info, servers)
-        paths = ""
-        paths_data = @document.paths
-        paths_data.each do |path_data|
-          # For some reason paths.each returns an array of arrays [title, object]
-          # instead of an array of objects
-          text = path_data[0]
-          paths += path(text)
+        paths = @document.paths.keys.inject("") do |memo, text|
+          memo + path(text)
         end
-        schemas = ""
-        schemas_data = @document.components.schemas
-        schemas_data.each do |schema_data|
-          text = schema_data[0]
-          schemas += schema(text)
+
+        schema_names = @document.components.schemas.keys
+        schemas = schema_names.inject("") do |memo, schema_name|
+          memo + schema(schema_name)
         end
+
         @template_api_full.result(binding)
       end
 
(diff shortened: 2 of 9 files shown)


lib/govuk_tech_docs/api_reference/api_reference_renderer.rb
```diff
@@ -167,108 +123,52 @@ def json_prettyprint(data)
...
-          all_of = schema["allOf"]
+        properties.each_with_object({}) do |(name, schema), memo|
+          memo[name] = if schema.type == 'object'
+                         if schema.properties
+                           schema_properties(schema)
+                         elsif schema.items
+                           schema_properties(schema.items)
+                         else
+                           {}
+                         end
```
ChrisBAshton: This block feels like it could be refactored - something like:

```ruby
memo[name] = schema_properties(schema.items || schema)
memo[name] = (schema.example || schema.type) unless memo[name]
```

But I might be missing a nuance there - and appreciate that you're concerned about introducing new bugs.

lib/govuk_tech_docs/api_reference/api_reference_renderer.rb
```diff
@@ -167,108 +123,52 @@ def json_prettyprint(data)
...
-
-        # Schema dictates that it's always components['schemas']
-        text.gsub(/#\/components\/schemas\//, '')
+        {
+          'get' => path.get,
+          'put' => path.put,
+          'post' => path.post,
+          'delete' => path.delete,
+          'patch' => path.patch,
+        }.compact
```
ChrisBAshton: 🎉

CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # Changelog
 
-## 2.0.5
+## Unreleased
+
+Fix bug in API docs that shows required attributes as optional (#113)
```
ChrisBAshton: Code looks good in this commit - but can you fix the URL in the commit message so that it's not split onto multiple lines?

tijmenb: Sorry @kevindew, closed #113 because I had misunderstood the relationship between JSON Schema and OpenAPI.

lib/govuk_tech_docs/api_reference/api_reference_renderer.rb
kevindew: Thanks - I've not been very happy with my changes there. I don't dare quite go to the level you suggest here as there are a few more nuances. I've done a bit of refactoring to try make it all a bit easier to read though.

lib/govuk_tech_docs/api_reference/api_reference_renderer.rb
```diff
@@ -18,20 +18,15 @@ def initialize(app, document)
...
 
       def api_full(info, servers)
-        paths = ''
-        paths_data = @document.paths
-        paths_data.each do |path_data|
-          # For some reason paths.each returns an array of arrays [title, object]
-          # instead of an array of objects
-          text = path_data[0]
-          paths += path(text)
+        paths = @document.paths.keys.inject('') do |memo, text|
```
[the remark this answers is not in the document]
kevindew: `#inject` pre-dates `#reduce` in Ruby so that seems the more idiomatic way to do it for me. I wouldn't suggest it needs changing here as we've never applied any preference in GDS code for one way or the other.

CHANGELOG.md
kevindew: Sure I've fixed that.

kevindew:
@ChrisBAshton Thanks for the review and apologies in taking so long to get back to this. I've fixed up various bits noted.

lib/govuk_tech_docs/api_reference/templates/schema.html.erb
```diff
@@ -11,7 +11,7 @@
 <td><%= property[0] %></td>
 <td><%= property[1].type %></td>
 <td><%= property[1].required.present? %></td>
-<td><%= markdown(property[1].description) %></td>
+<td><%= property[1].description_html %></td>
```
ChrisBAshton: While you're here, could you apply the same changes to this file as you did in the "Rename some variables for clarity" commit?

CHANGELOG.md
ChrisBAshton: Think we ought to remove this line from this commit now that it's been closed without merging. (Should mean the last commit isn't needed either).

lib/govuk_tech_docs/api_reference/templates/schema.html.erb
kevindew: I don't think it's necessary to do a change here as it's done in Tijmen's commit which is next.

kevindew:
@ChrisBAshton about 1 million years later, but this is good for a re-review if you don't mind? 🙏

lib/govuk_tech_docs/api_reference/templates/schema.html.erb
```diff
@@ -6,18 +6,18 @@
 <tr><th>Name</th><th>Type</th><th>Required</th><th>Description</th><th>Schema</th></tr>
 </thead>
 <tbody>
-<% properties.each do |property| %>
+<% properties.each do |property_name, property_attributes| %>
```
ChrisBAshton, line 9: Minor: the commit message confusingly refers to a `attribute_name`, I assume it should refer to `property_name` instead.

spec/features/api_reference_spec.rb
```diff
@@ -55,5 +55,7 @@ def then_there_is_correct_api_schema_content
     expect(page).to have_css("h3#schema-pet", text: "Pet")
     # Schema parameters
     expect(page).to have_css("table", text: /\b(tag )\b/)
+    # Check that the "required" column is true for the `id` attribute
+    expect(page).to have_css("table.schema-pet td:nth(3)", text: "true")
```
ChrisBAshton, line 59: Would be great if this test were a bit more self-documenting - I wonder if we can dynamically infer the correct column number? I appreciate that introduces some complexity to the test, so may not be an improvement.

lib/govuk_tech_docs/api_reference/templates/api_reference_full.html.erb
```diff
@@ -1,7 +1,8 @@
 <h1 id="<%=  info.title.parameterize %>"><%= info.title %> v<%= info.version %></h1>
 <%= info.description_html %>
 
-<% unless servers.empty? %>
+<%# OpenAPI files default to having a single server of URL "/" %>
```
ChrisBAshton, line 4:
```suggestion
<% # OpenAPI files default to having a single server of URL "/" %>
```

ChrisBAshton (approved):
LGTM 👍 some non-blocking comments below.

lib/govuk_tech_docs/api_reference/templates/schema.html.erb
kevindew: Sure - I've changed it (felt a bit weird editing someone else's commit)

spec/features/api_reference_spec.rb
kevindew: I don't really want to edit someone else's code there, nor particularly going into the world of testing for the API stuff. It seems a big can of worms as there is a lack of testing for it all.

lib/govuk_tech_docs/api_reference/templates/api_reference_full.html.erb
kevindew: `<%#` is actually erb syntax for a comment and means something different to `<% #`: https://docs.ruby-lang.org/en/2.3.0/ERB.html#class-ERB-label-Recognized+Tags

ChrisBAshton: Ooh, TIL! 👀

PR #84, opened 2019-04-09
fix: generate multipage TOC with http_prefix

lhokktyn:
Please see replication steps here: https://github.com/alphagov/tech-docs-template/issues/183#issuecomment-478942982

Whilst the test already in place appears to work, I'm not sure it is covering my particular case because when I build my site it is not generating files within a prefixed folder, so I may need a bit of guidance to write additional test (or correct my approach!) ...

For example, when `http_prefix=/prefix/`, given my `source/`:

```
|- source/
|--documentation/
|--index.html.md.erb
|--page2.html.md.erb
```

This is generating files in the build directory as so:

```
|- build/
|--index.html
|--page2.html
```

But the test in place appears to expect generated assets to be:

```
|- build/
|--prefix/index.html
|--prefix/page2.html
```

The attached PR does fix the issue for me, however.

PR #136, opened 2019-10-31
Update multi-page navigation to allow custom path prefix

### LINKED ISSUE #128: Navigation is broken when deploying on a path

alex-ju:
Currently if the tech-docs is deployed on a path (e.g. `https://alphagov.github.io/api-catalogue/`) the navigation links are broken as they are generated from root (`/`). One way to avoid this without affecting current single-page or multi-page configuration is to define a `path_prefix` in config that is prepended to each link.

This is an attempt to fix #128 which is currently blocking any deployment for the [api-catalogue](https://github.com/alphagov/api-catalogue).

I'm open to any other suggestions that will unblock this - especially since I'm not familiar with this project or Middleman.

Note: I needed to refactor the conditional statements as linter was complaining about the code having too many lines (`Metrics/BlockLength: Block has too many lines. [31/25]`)

### DIFF
--- a/lib/govuk_tech_docs/table_of_contents/helpers.rb
+++ b/lib/govuk_tech_docs/table_of_contents/helpers.rb
@@ -55,22 +55,18 @@ module GovukTechDocs
           # prefix in to consideration when checking for the root index.html.
           # The prefix may be set with or without a trailing slash: make sure
           # it has one for this comparison check.
-          home_url =
-            if config[:http_prefix].end_with?("/")
-              config[:http_prefix]
-            else
-              config[:http_prefix] + "/"
-            end
+          home_url = config[:http_prefix].end_with?("/") ? config[:http_prefix] : config[:http_prefix] + "/"
+          resource_url = config[:tech_docs][:path_prefix] ? config[:tech_docs][:path_prefix] + resource.url : resource.url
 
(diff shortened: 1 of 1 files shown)


alex-ju:
> Could we add some tests for this?

I thought a lot about that, but not sure where it fits as the test suite currently relies on one configuration file. This change will require a separate config and not sure where to accommodate that. Open to suggestions.

alex-ju:
> I don't see how this differs to `http_prefix` 🤔 and if `http_prefix` isn't doing what we want, it's probably a bug in its implementation and we should fix that rather than add a new property.

Ha, I haven't thought about that - I saw it as more of a way to specify the protocol (http/https). I'll give it a go, if it doesn't work I would really appreciate a hand, thank you!

ChrisBAshton:
I agree the naming could be better!

alex-ju:
As mentioned in the comments – this seems to have been addressed already.

alex-ju:
This is still a problem – maybe https://github.com/alphagov/tech-docs-gem/pull/84 will fix it.

PR #151, opened 2019-11-11
Add signposting after moving `docs` folder content

bravegrape:
## What

There's a PR to move the content from `docs` folder to `alphagov/tdt-documentation`: https://github.com/alphagov/tdt-documentation/pull/108

The move is to make it easier to work on improving the documentation. Raising PRs for changes across 2 repos is annoying. The long-term plan is to move the docs to the gem itself, but we need to work out some technicalities first.

This PR is to:

* remove the content of the `docs` folder if https://github.com/alphagov/tdt-documentation/pull/108 is merged
* add appropriate signposting about where the content from the `docs` folder went.

## Why

This PR aims to avoid content duplication and to not confuse contributors to the tech-docs gem.

The CONTRIBUTING.md file talks about updating content in the `docs` folder so we should update this to the new location for documentation. We should also say where the content of `docs` went.

## How to review

Check that signposting is sensible.
Only merge if https://github.com/alphagov/tdt-documentation/pull/108 is merged.


36degrees:
> The long-term plan is to move the docs to the gem itself, but we need to work out some technicalities first.

Just to check my understanding, does that mean we're moving the documentation out of this repo, but at some point in the future we plan to bring it back?

docs/configuration.md
```diff
@@ -1,228 +1,3 @@
...
-Decides whether or not to display the page review banner, regardless of whether the page needs to be reviewed or not.
-
-If not present or set to `true`, the banner will be displayed on the page. This is the default behaviour.
-
-If set to `false`, the banner will not be displayed.
-
-See the separate [documentation for page expiry][expiry] for more details.
-
-[expiry]: https://tdt-documentation.london.cloudapps.digital/page-expiry.html#page-expiry-and-review
+Content has been moved to alphagov/tdt-documentation: https://github.com/alphagov/tdt-documentation/pull/108/commits/e50c6bc0dd43b2f299444b7dc760e1a85edcb718
```
36degrees: Would it be more useful to link the user to the live documentation site for these links (e.g. https://tdt-documentation.london.cloudapps.digital/amend_project/configuration/#configuration-options)?

bravegrape: My reasoning for putting that link in was to explain why it was moved. But I think you're right, it's probably more helpful to tell users where the content was moved to.

How about changing it to something like:

This content has been moved to: https://tdt-documentation.london.cloudapps.digital/amend_project/configuration/#configuration-options

So you think it's worth linking to the source file too?

bravegrape:
> > The long-term plan is to move the docs to the gem itself, but we need to work out some technicalities first.
> 
> Just to check my understanding, does that mean we're moving the documentation out of this repo, but at some point in the future we plan to bring it back?

@36degrees yup, that's been the plan for a while now (https://github.com/alphagov/tech-docs-gem/pull/83) but it's harder to find the time to do it. I went for the moving back and forth approach to allow progress in smaller chunks of work, rather than snowballing the migration work.

docs/configuration.md
bravegrape: Added links to where the content is published: https://github.com/alphagov/tech-docs-gem/pull/151/commits/244142097cc6f5b85ba4ab5c6f0674dc2c690c53

The pages have links to the `alphagov/tdt-documentation` repo as well as the contribution banner so it should be fairly straightforward to find the source.

Issue #133, opened 2019-10-28
Columns in tables now display with equal widths


klssmith:
The width of the columns in a table used to vary according to the content, but now all the columns have the same width, which can make some content appear quite squashed. @nickcolley pointed out that it seems to be related to this change https://github.com/alphagov/tech-docs-gem/commit/ffa22615851866bb7052a87285fca30273f0ac54

**Before**
![Screen Shot 2019-09-10 at 14 44 41](https://user-images.githubusercontent.com/12881990/67081892-fc7a4680-f18f-11e9-89b8-90b4608e27b9.png)

**After**
![Screen Shot 2019-09-10 at 14 44 50](https://user-images.githubusercontent.com/12881990/67081915-07cd7200-f190-11e9-8191-98ab28725a72.png)

tombye:
Fixed in https://github.com/alphagov/tech-docs-gem/pull/154

![image](https://user-images.githubusercontent.com/87140/68768317-30734b00-061a-11ea-9b66-e6308bbd3dcc.png)

Issue #100, opened 2019-09-04
Code block border does not match code block area


jonathanglassman:
For example:

https://govukpay-docs-test-jfh.cloudapps.digital/making_payments/#creating-a-payment

![Making payments   Pay](https://user-images.githubusercontent.com/29436951/64270953-59010980-cf34-11e9-9f25-fa060668351a.png)

jonheslop:
Oh weird I didnt understand what you meant by this! What browser were you using because it doesn’t do this in Firefox, which is why I was confused

jonathanglassman:
> Oh weird I didnt understand what you meant by this! What browser were you using because it doesn’t do this in Firefox, which is why I was confused

@jonheslop this happens in Chrome

bravegrape:
@jonheslop also in Safari.

This reminds me of something @christophertomasdesign noticed about the tech docs template when he made a diagram for the docs. He said the text width in the tech docs template is around 100 px  (?) larger than the text width on GOV.UK. Maybe the two are related?

jonheslop:
This issue is caused by 
https://github.com/alphagov/tech-docs-gem/blob/master/lib/assets/stylesheets/modules/_technical-documentation.scss#L122-L126

Which is seems like was a temp fix for another issue. I was just gonna remove the `max-width: 40em` until I saw the comment 🤔 

But @bravegrape you’re probably right that the line length in general is probably too long

PR #154, opened 2019-11-12
Reset `table-layout` to `auto`

tombye:
This removes the `table-layout: fixed` style for tables in tech' docs content, introduced as part of https://github.com/alphagov/tech-docs-gem/commit/ffa22615851866bb7052a87285fca30273f0ac54. 

## Why did we fix our table layouts?

It was introduced as part of a pull request that made 2 changes:
1. the `max-width: 40em` for `<pre>` blocks was removed to stop [issue #100](https://github.com/alphagov/tech-docs-gem/issues/100)
2. `table-layout: fixed` was added to fix the problems `max-width: 40em` solved

## Why did we constrain the width of our `<pre>` tags?

The comments above `max-width: 40em` mention `<pre>` tags pushing tables to be larger than the main column (see https://github.com/alphagov/tech-docs-gem/commit/ffa22615851866bb7052a87285fca30273f0ac54#diff-d7f12a4697fd2195a8c62339caee5c77L122).

### 1. It causes an issue with existing tables

It created [issue #133](https://github.com/alphagov/tech-docs-gem/issues/133) with Notify's tech' docs.

### 2. Removing it shouldn't effect any other existing tables

I went through the main users of this gem and found all the tables in them (thanks @nickcolley for their details):

https://docs.google.com/spreadsheets/d/1OZGQifZIw0I2GXqmrp78DoI3R2sfcxn5IcFq3Y5RXqA/edit#gid=0

I can't find any tables that use `<pre>` tags in table cells, just inline `<code>` blocks.

I also tested all these tables without `table-layout: fixed` and none of them displayed in ways that distorted the content or made the table wider than the main column.

### 3. Tables that are too wide don't break the main column

Tables are wrapped in a fixed-width container which scrolls. This means that if they're wider than the main column they don't break it and users can scroll to access their content. For example:

https://www.docs.verify.service.gov.uk/legacy/build-ms/msa/#msa-tls-certificates

### We probably shouldn't be using block-level code in tables

Note: this point is just my opinion 🙂.

Middleman's markdown parser [Redcarpet uses PHP-Markdown style tables](https://github.com/vmg/redcarpet#and-its-like-really-simple-to-use. In its docs, PHP-Markdown says it allows 'span level formatting' in its tables:

https://michelf.ca/projects/php-markdown/extra/#table

There's no mention of official support for block-level formatting, like `<pre>`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,7 @@
 
 ## Unreleased
 
-Remove the `table-layout: fixed` style from the tables in tech' docs content, mainly due to it
-causing https://github.com/alphagov/tech-docs-gem/issues/133.
+Change how tables are laid out to fix https://github.com/alphagov/tech-docs-gem/issues/133.
 
 ## 2.0.8
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 ## Unreleased
 
-Change how tables are laid out to fix https://github.com/alphagov/tech-docs-gem/issues/133.
+Change how tables are laid out to fix some content in Notify tech' docs being squashed due to its
+column having a fixed width.
+
+Related issue: https://github.com/alphagov/tech-docs-gem/issues/133
 
 ## 2.0.8
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+Remove the `table-layout: fixed` style from the tables in tech' docs content, mainly due to it
+causing https://github.com/alphagov/tech-docs-gem/issues/133.
+
 ## 2.0.8
 
 Use of  `govuk-lint` replaced with `rubocop-govuk` due to the former [becoming deprecated](https://github.com/alphagov/govuk-lint/pull/133).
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+Remove the `table-layout: fixed` style from the tables in tech' docs content, mainly due to it
```
nickcolley: Not blocking by ideally the changelog should say what is fixed for users, not the details of what we've been doing internally.

tombye: Makes sense. Does [this change](https://github.com/alphagov/tech-docs-gem/pull/154/commits/1f90f78c00bbe7f1426aacc476dcde1325284db3) help?

nickcolley: Great, I'd prefer if it didnt require looking in the link to understand the summary but good enough 👍 Thanks Tom

tombye: That's a good point actually. I've rewritten it in [this commit](https://github.com/alphagov/tech-docs-gem/pull/154/commits/a4a20f19eaf412c8e455768f974eb824f1374d8d).

Issue #110, opened 2019-09-13
Zooming and scaling does not work properly on mobile devices


jakubmiarka:
The scaling and zooming on mobile devices is disabled:
https://github.com/alphagov/tech-docs-gem/blob/6c44ea2c6e8787046af0fe4fa6049c9609f2beab/lib/source/layouts/core.erb#L6

This fails [WCAG 2.1 - 1.4.4 Resize text Level AA criterion](https://www.w3.org/WAI/WCAG21/quickref/#resize-text).

36degrees:
From memory, this is because of the sticky 'table of contents' on mobile, and the peculiar way that sticky elements scale and position themselves when zooming (I think there may also have been issues with the table of contents overlay that appears)

As a lot of the tech docs using this format are now split across multiple pages, it may be that the table of contents overlay can be redesigned or removed, but this isn't a trivial change and IMHO needs to be designed and researched.

nickcolley:
Testing this on Android Chrome, it lets you zoom regardless of if we've turned it off.

Will need to test on more devices to see what the behaviour is.

PR #155, opened 2019-11-12
Ensure the page can be pinch to zoomed

nickcolley:
We have intentionally tried to disable pinch to zoom in the past based around the sticky navigation, but this is an accessibility issue.
This pull request updates this decision to pinch to zooming, which for the most part is how it's working at the moment anyway since a lot of devices ignore blocking pinch to zoom.

I've tested in latest versions of iOS and Android.

- iOS v10 onwards ignores disabling pinch to zoom
- Android does not ignore disabling pinch to zoom, although the user can override this.

I have not been able to introduce any negative impact for allowing pinch to zoom, the interface scales proportionally and does not obstruct anything. So given that iOS already ignores this I think we should allow this for all users.

This is also a WCAG accessibility requirement.

Fixes https://github.com/alphagov/tech-docs-gem/issues/110

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,9 @@
 
 ## Unreleased
 
-Change how tables are laid out to fix some content in Notify tech' docs being squashed due to its
-column having a fixed width.
+### Fixes
 
-Related issue: https://github.com/alphagov/tech-docs-gem/issues/133
+- [Pull request #154: Fix table columns breaking content](https://github.com/alphagov/tech-docs-gem/pull/154).
 
 ## 2.0.8
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixes
 
 - [Pull request #154: Fix table columns breaking content](https://github.com/alphagov/tech-docs-gem/pull/154).
+- [Pull request #155: Ensure the page can be pinch to zoomed](https://github.com/alphagov/tech-docs-gem/pull/155).
 
 ## 2.0.8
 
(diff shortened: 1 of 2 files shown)


nickcolley:
I tested this with browserstack, which allows pinch to zoom gestures via a trackpad.

@36degrees @hannalaakso I would appreciate you double checking my findings if you have the time.

hannalaakso:
Results of my testing in the lab:
Safari on iOS and Samsung Internet on Android looks good. 
Chrome on iOS and Firefox on Android don't seen to work.

nickcolley:
@hannalaakso can you give more detail on 1. what URL you were testing 2. what happened to make it seem broken. This change uses the same meta tag as the Design System so I'm confused why pinch to zoom wouldnt work with this change.

hannalaakso:
Sorry for the terseness, I was rushing off into a meeting 😞 

I tested the pinch to zoom on https://tdt-documentation.london.cloudapps.digital.

I've since managed to install Chrome on the new iOS devices in the lab and pinch to zoom works on those so the issue on iPhone 5s could be a hardware issue on older iPhones or something to do with iOS 12 (although @nickcolley's findings above suggest this shouldn't be an issue > iOS 10 - I'll try to find another iOS device to test on if we want to dig into this further).

✅ Safari on iPhone 5s 12.1
❌ Chrome on iPhone 5s 12.1 *
✅ Safari on iPhone 8 13.2
✅ Chrome on iPhone 8 13.2
✅ Chrome on iPad 3 13.2
✅ Safari on iPad 3 13.2
✅ Samsung Internet 10.1 on Galaxy S8
❌ Firefox 68 on on Galaxy S8 *
 
* Pinch to zoom works on https://design-system.service.gov.uk/

nickcolley:
@hannalaakso I think those findings confirm what I found, since this pull request has not been deployed yet I would expect some devices to not allow pinch to zoom right now.

nickcolley:
OK I'll add a changelog entry and we can get this merged, thanks for the extensive testing Hanna

PR #160, opened 2019-11-14
Make sure IDs on collapsible navigation are unique

jakubmiarka:
Multiple nested pages were getting assigned the same ID (from the parent). This change
appends the iterator index to the elements to make sure they're unique.
Also added an extra nested page to the example docs and a test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- [Pull request #160: Make sure IDs on collapsible navigation are unique](https://github.com/alphagov/tech-docs-gem/pull/160)
+
 ## 2.0.9
 
 ### Fixes
--- a/example/source/nested-page/another-nested-nested-page/index.html.md
+++ b/example/source/nested-page/another-nested-nested-page/index.html.md
@@ -1,0 +1,5 @@
+---
+title: Another nested nested page
+---
+
+# Another nested nested page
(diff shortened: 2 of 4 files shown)


nickcolley:
Looks like there's some issues with the format of the JavaScript as StandardJs is throwing errors.

I will check this out but seems sensible so far, thanks :)

jakubmiarka:
> Hmm actually this seems to break the relationship between aria-controls on the button and the expanded element. 🤔

@nickcolley I have amended the the logic in https://github.com/alphagov/tech-docs-gem/pull/160/commits/41c09567950edb75f86aae726fece3c7f9f16508 ... the other option was to wrap the headings in a single element for aria purposes, thought this was easier. Let me know what you think.

nickcolley:
I imagine from a real world point of view having a single expanded area would be easier to use. I suspect having multiple places being controlled may behave in a weird way in screen readers even though it's allowed by the aria specification.

nickcolley:
Maybe worth giving this a test in a screen reader before making a decision? It may work great with multiple IDs.

jakubmiarka:
From what I can see, `aria-controls` is [only supported](https://a11ysupport.io/tests/tech__aria__aria_controls) in JAWS and supposedly* it just says "press the JAWS key plus Alt and M to move to the controlled element" when using multiple IDs. 
So we can either leave it as it as and be compliant with the spec, or add a wrapper around the expandable area. However, this will be a slightly more work as there are inconsistencies how the tree is getting build using headers vs nested pages.

*I don't have the capacity to be testing this with JAWS right now

nickcolley:
I think with that in mind we should get this in and consider raising an issue to improve it in the future.

PR #162, opened 2019-11-15
Fix line height spacing for multiline code elements

tombye:
Text in our `<code>` elements is larger than the default Transport font on OSX, due to it using the Monaco font. This creates a taller formatting box for each line of text it has.

We set a smaller font-size, and line-height in `<td>`s. This works for Transport but text in `<code>` sections appears squashed vertically.

This sets the line-height for `<code>` to match that of content outside tables, to stop it
squashing up.

This has been tested with Consolas (the font Windows uses for `<code>`) and all other
fallbacks.

Here's an idea of the changes this introduces, viewed on OSX Firefox.

### Before

image

## After

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,14 @@
 
 ## Unreleased
 
+### Fixes
+
+- [Pull request #162: Fix line height spacing for multiline code elements](https://github.com/alphagov/tech-docs-gem/pull/162)
+- [Pull request #165: Update header alignment to match layout](https://github.com/alphagov/tech-docs-gem/pull/165)
+
 ## 2.0.10
+
+### Fixes
 
 - [Pull request #160: Make sure IDs on collapsible navigation are unique](https://github.com/alphagov/tech-docs-gem/pull/160)
 
--- a/lib/assets/stylesheets/modules/_technical-documentation.scss
+++ b/lib/assets/stylesheets/modules/_technical-documentation.scss
@@ -139,6 +139,8 @@
     font-family: monaco, Consolas, "Lucida Console", monospace;
     font-size: 15px;
     font-size: 0.9375rem;
+    // Match the line-height outside of tables
+    line-height: 1.4;
     color: $code-0E;
 
     @include govuk-media-query(tablet) {
--- a/example/source/code.html.md
+++ b/example/source/code.html.md
@@ -8,6 +8,35 @@ A paragraph with a `code` element within it.
 
 <a href="#"><code>code element within a link</code></a>
 
+An example of a table with a `code` element within it.
+
+<div class="table-container">
+  <table>
+    <thead>
+      <tr>
+        <th style="text-align:left">httpResult</th>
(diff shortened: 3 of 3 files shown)


lib/assets/stylesheets/modules/_technical-documentation.scss
```diff
@@ -141,6 +141,8 @@
     font-family: monaco, Consolas, "Lucida Console", monospace;
     font-size: 15px;
     font-size: 0.9375rem;
+    // Match the line-height outside of tables
+    line-height: 1.4;
```
nickcolley: I wonder if the keyword `normal` would work, but happy with this if not.

nickcolley:
I wonder if this is related: https://github.com/alphagov/tech-docs-gem/pull/131

lib/assets/stylesheets/modules/_technical-documentation.scss
tombye: I tested this and it would work for all the monospaced fonts we specify but the padding on the top and bottom of the line still causes some vertical overlap.

tombye:
> What impact does this have to code elements within paragraphs?

It seems to work OK:

image
(from Notify's REST API docs.)

Looks like the line-height leaves enough space for the padding on the top and bottom.

nickcolley:
I'm still wondering if the reason this has broken is because the relative font units is not working in that area for some reason 🤔

nickcolley:
Would you mind adding an example to: http://127.0.0.1:4567/code.html#code-examples please?

tombye:
> I'm still wondering if the reason this has broken is because the relative font units is not working in that area for some reason 🤔
> 
> The text looks a lot bigger than the surrounding regular text.

On smaller screens, the `<code>` **is** larger than the normal table text, 15px vs 14px, whatever the monospace font-family. On larger screens, they're both 16px.

I did some testing on the differences in vertical size between the monospace font-families and New Transport.

I measured the height from the ascender to the descender on various combinations, all at 16px, and found:
- Monaco and New Transport at 16px were roughly the same
- Consolas and New Transport showed some difference with Consolas being about 3px smaller
- monospace was the same as Consolas

All was on Firefox on OSX.

nickcolley:
I'll try and review this when I get some free time, please ping me if I don't.

tombye:
> I'll try and review this when I get some free time, please ping me if I don't.

Will do 👍

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+## 2.0.10
```
nickcolley:
```suggestion
### Fixes
```

nickcolley: This needs rebasing and adding to the 'fixes' section of the unreleased stuff

nickcolley: Happy to do this for you if you want

nickcolley: Tom said he was happy for me to get this sorted

PR #165, opened 2019-11-19
Update header alignment to match layout

nickcolley:
This pull request aims to address some spacing issues introduced when this format was updated to use the GOV.UK Design System by aligning the header and layout elements on the page.

It also expands the width of the area for the product name which stops it wrapping early, this is something we might want to consider pushing upstream to GOV.UK Frontend in the future.

| Full size before | Full size after |
| - | - |
|  |  |

| Smaller before | Smaller after |
| - | - |
|  |  |

| Even smaller before | Even smaller after |
| - | - |
|  |  |

Fixes https://github.com/alphagov/tech-docs-gem/issues/164

### DIFF
--- a/lib/assets/stylesheets/_core.scss
+++ b/lib/assets/stylesheets/_core.scss
@@ -6,6 +6,7 @@
 @import "modules/anchored-heading";
 @import "modules/app-pane";
 @import "modules/footer";
+@import "modules/header";
 @import "modules/page-review";
 @import "modules/search";
 @import "modules/contribution-banner";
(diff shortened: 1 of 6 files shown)


36degrees:
> It also expands the width of the area for the product name which stops it wrapping early, this is something we might want to consider pushing upstream to GOV.UK Frontend in the future.

Possibly related: https://github.com/alphagov/govuk-frontend/issues/1502

nickcolley:
I've made a small tweak to align this even more:

| Before | After |
| - | - |
|  |  |

nickcolley:
@alex-ju if you could help us document where you've done similar things by commenting upstream here: https://github.com/alphagov/govuk-frontend/issues/1502 that'd be really helpful.

nickcolley:
I've asked the graphic designers if I could get an extra pair of eyes on this so will see what they think.

nickcolley:
I think this needs slightly adjusting for when there's no sidebar, so will make those updates...

nickcolley:
Laurence B and Charlotte D from the GDS graphic designer community have said they approve of this change so I'm going to get it merged.

Issue #169, opened 2020-02-24
Intent to remove: indentation turning markdown into code blocks.


nickcolley:
I think we should turn off the feature to allow indented code blocks to allow for more complex readable views.

## Fenced code block

```markdown
` ` `javascript
function helloWorld () {
    console.log("Hello, World.")
}
helloWorld()
` ` `
```

## Indented code block
```markdown
    function helloWorld () {
       console.log("Hello, World.")
    }
    helloWorld()
```

When working on more complex pages it's good to be able to indent your code within conditional statements so it's readable.

```markdown
    <% if condition %>
        ## Conditional section

        I'm only here some of the time.
    <% end %>
```

However by doing so the markdown renderer will turn code into code blocks, which means you're forced to have inconsistent indentation.

```markdown
    <% if condition %>
## Conditional section

I'm only here some of the time.
    <% end %>
```

We use fenced codeblocks mainly for the ability to be able to set the language of the code and allow for syntax highlighting, so I think it's good to always be explicit.

We can use the `disable_indented_code_blocks` option for the markdown renderer to do this.

nickcolley:
We may revisit this, closing for now.

PR #170, opened 2020-02-24
Allow additional markdown options

NickColley:
I think turning on new features may be too controversial so I'm adding functionality so the user can set their own preferences.

See https://github.com/alphagov/tech-docs-gem/issues/169 for why I want to do this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Fixes
+
+- [Pull request #162: Fix line height spacing for multiline code elements](https://github.com/alphagov/tech-docs-gem/pull/162)
+- [Pull request #165: Update header alignment to match layout](https://github.com/alphagov/tech-docs-gem/pull/165)
+
 ## 2.0.11
 
 ### Fixes
--- a/lib/govuk_tech_docs.rb
+++ b/lib/govuk_tech_docs.rb
@@ -40,15 +40,19 @@ module GovukTechDocs
     context.files.watch :source, path: "#{__dir__}/source"
 
     context.set :markdown_engine, :redcarpet
-    context.set :markdown,
-                renderer: TechDocsHTMLRenderer.new(
-                  with_toc_data: true,
-                  api: true,
-                  context: context,
-                ),
-                fenced_code_blocks: true,
-                tables: true,
-                no_intra_emphasis: true
+
+    default_markdown_options = {
+      renderer: TechDocsHTMLRenderer.new(
+        with_toc_data: true,
+        api: true,
+        context: context,
+      ),
+      fenced_code_blocks: true,
+      tables: true,
+      no_intra_emphasis: true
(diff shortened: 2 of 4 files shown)


nickcolley:
Can we do this via the config.yaml file?

lib/govuk_tech_docs.rb
```diff
@@ -40,15 +43,19 @@ def self.configure(context, options = {})
...
+      renderer: TechDocsHTMLRenderer.new(
+        with_toc_data: true,
+        api: true,
+        context: context,
+      ),
+      fenced_code_blocks: true,
+      tables: true,
+      no_intra_emphasis: true
+    }
+    markdown_options = context.config[:tech_docs][:markdown].transform_keys(&:to_sym) || {}
```
36degrees: What happens if there are no markdown options specified in the config file? Does this need to be guarded?

nickcolley: I believe It'll return falsy and the {} will be used.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Fixes
```
nickcolley: this was accidentally added...

lib/govuk_tech_docs.rb
36degrees: Doesn't seem to like the chaining when I try it:

```
irb(main):004:0> foo = { :tech_docs => { :something => true }}
=> {:tech_docs=>{:something=>true}}
irb(main):005:0> foo[:tech_docs][:markdown].transform_keys(&:to_sym) || {}
Traceback (most recent call last):
        4: from /Users/oliverbyford/.rbenv/versions/2.6.3/bin/irb:23:in `<main>'
        3: from /Users/oliverbyford/.rbenv/versions/2.6.3/bin/irb:23:in `load'
        2: from /Users/oliverbyford/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        1: from (irb):5
NoMethodError (undefined method `transform_keys' for nil:NilClass)
```

Looks like it works OK with safe navigation (`&.`) though:

```
irb(main):004:0> foo = { :tech_docs => { :something => true }}
=> {:tech_docs=>{:something=>true}}
irb(main):006:0> foo[:tech_docs][:markdown]&.transform_keys(&:to_sym) || {}
=> {}
irb(main):007:0> foo = { :tech_docs => { :something => true, :markdown => { :be_awesome => true }}}
=> {:tech_docs=>{:something=>true, :markdown=>{:be_awesome=>true}}}
irb(main):008:0> foo[:tech_docs][:markdown]&.transform_keys(&:to_sym) || {}
=> {:be_awesome=>true}
```

nickcolley:
Gonna pause this, write some cards and we can revisit if it makes sense.

PR #176, opened 2020-04-16
Use correct separator for page titles

36degrees:
Use a hyphen (`-`) rather than a pipe (`|`) within page titles to be consistent with [the guidance in the Service Manual](https://www.gov.uk/service-manual/design/writing-for-user-interfaces#headings-and-title):

> The <title> should be based on the `<h1>`, and follow this format:
>
> Where do you live? - register to vote - GOV.UK


36degrees:
Should pass if rebased once #178 has been merged.

PR #192, opened 2020-08-14
Fixes and improvements to meta tags

36degrees:
- Fix invalid meta tags (~using invalid `property` attribute rather than `name`~ EDIT: non opengraph tags using the non-standard `property` attribute, which should only be used for opengraph / RDFa tags)
- Move robots and Google site verification meta tag logic to the meta_tags helper
- Make it possible to exclude individual pages from search engine indexes by setting `prevent_indexing: true` in the page frontmatter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Changelog
 
+## Unreleased
+
+### New features
+
+#### Exclude individual pages from search engine indexes
+
+You can now exclude individual pages from search engine indexes by including `prevent_indexing: true` in the frontmatter for the page.
+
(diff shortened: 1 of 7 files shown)


36degrees:
Realised that OpenGraph meta tags do in fact use the non-standard `property` attribute. I think that means the non-opengraph `description` meta tag is incorrect, but the others were actually correct before.

Will need to work out how to make the meta tags helper handle both cases.

36degrees:
I've updated the meta tags helper to return opengraph and non-opengraph tags separately, which I think solves the issue mentioned in the previous comment…

freesteph (approved):
Looks good to me :+1:

PR #193, opened 2020-09-04
Release v2.1.0

36degrees:


kr8n3r (approved):

m-green:
Release note looks good! Couple of suggestions.

- There's a mix between plural and singular in the first paragraph ('You can now exclude individual pages' vs 'for the page').  Best to stick with singular, so probably better to say 'You can now exclude a page'.
- I'd probably remove 'indexes' and just say 'from search engines'.
- Do we need to mention 'to the Tech Docs Gem' specifically in the fixes section? (If we do, I'm not sure about the capitals...)

PR #198, opened 2020-10-01
Cache gems + node modules on CI

odlp:
Whilst contributing other PRs I noticed the time to run tests on CI could be improved by using the [Travis dependency cache](https://docs.travis-ci.com/user/caching#caching-directories-bundler-dependencies).

Before the change, recent builds on `master` have taken approx 5m30s-5m50s to complete - including over 3 minutes spent installing gems. With a warm cache this drops to ~2m30s build time, and installing gems takes under 10 seconds.

## Before

build-before

## After

build-after

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: ruby
+cache:
+  bundler: true
+  npm: true
 before_install:
   - gem update --system
   - gem install bundler


.travis.yml
```diff
@@ -1,4 +1,7 @@
 language: ruby
+cache:
+  bundler: true
+  npm: true
```
odlp, line 4: Travis already caches NPM install / node modules automatically, but I've made this explicit in the config so readers don't need to know about this default, and to maintain parity with the bundler caching.

richardTowers (approved):

PR #199, opened 2020-10-16
Fix page expiry box link colours (hover and focus)

danacotoran:
Adjust page expiry links' styles with the aim of resolving insufficient colour contrast issues on the hover and focus states.

## Issues ⬇️ 
Screenshot 2020-10-16 at 10 22 00
Screenshot 2020-10-16 at 10 22 13

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Fixes
+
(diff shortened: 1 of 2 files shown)


kr8n3r (approved):
💯

kr8n3r (changes requested):
focus state should have black text

kr8n3r (approved):

PR #203, opened 2021-02-03
Update jQuery from v1.12.4 to v3.5.1

chao-xian:
jQuery v3.5.1. available from:
https://code.jquery.com/jquery-3.5.1.min.js


ESKYoung:
Thanks for raising the draft PR @chao-xian! I've made an attempt at updating third-party JavaScript files. jQuery and Lodash are now up-to-date. Other points to note:

1. (lib/assets/javascripts/_vendor/modernizr.js) The existing Modernizr v3.3.1 file appears to be a custom build (see line 2). However, the link in line 2 to the Modernizr site should not be used, [as the site is out-of-date and broken](https://github.com/Modernizr/Modernizr#modernizr-is-a-javascript-library-that-detects-html5-and-css3-features-in-the-users-browser)
   - I tried rebuilding v3.3.1 using the same build parameters, but there appeared to be differences with our file
   - The latest version is v3.11.4, but I'm not sure how to make the custom build or if I should given the above
2. (spec/javascripts/helpers/lunr.min.js) I tried updating Lunr to v2.3.9, but the [tests failed](https://travis-ci.org/github/alphagov/tech-docs-gem/builds/757109116), and the built application.js file still had v0.7.0 in it, so I rolled back those changes
3. (lib/assets/javascripts/_vendor/jquery.mark.js) mark.js seems to be latest release, but the [repo](https://github.com/julmot/mark.js) seems to be stale (last master commit in Aug 2019)
4. (lib/assets/javascripts/_vendor/fixedsticky.js) Latest release of fixedsticky.js, but this [repo has been deprecated](https://github.com/filamentgroup/fixed-sticky).

PR #205, opened 2021-02-16
Improve table of contents accessibility

danacotoran:
The table of contents fails [WCAG 2.4.3](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html). 
When the table of contents is expanded, a search field and a list of (sometimes) collapsible options appear. Focus is automatically set to the first link on the list, which means that screen reader users might miss both the search box and the "hide table of contents" button.
The changes introduced in this PR ensure that focus is not set on the first nav option, but on the entire region. This way when the user navigates forward, they are able to encounter the search box and the "close" button.

Additionally, adjust the markup for the open/close TOC buttons to use the `<button>` element instead of an anchor tag. The reason is that these elements both look and behave as buttons, not as links. This should aid assistive tech users in navigating this functionality (for ex using voice control)

### DIFF
--- a/lib/assets/javascripts/_modules/table-of-contents.js
+++ b/lib/assets/javascripts/_modules/table-of-contents.js
@@ -7,23 +7,22 @@
     var $toc
     var $tocList
 
-    var $openLink
-    var $closeLink
+    var $openButton
+    var $closeButton
 
     this.start = function ($element) {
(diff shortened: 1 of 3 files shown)


kr8n3r (approved):
nice, good improvement. tested locally

PR #209, opened 2021-02-25
Some search and keyboard navigation updates

danacotoran:
Add a button to the search form as suggested in the DAC Accessibility audit. Make input label more descriptive to help screen reader users determine the purpose of the form.

### Search field before

Screenshot 2021-02-25 at 16 45 52

### Search field after

Screenshot 2021-02-25 at 16 45 29

Prevent tabbing beyond the TOC modal and/or search results when they are open. Screen reader users should also not be able to access page content that is not visible on the screen.

This should help address issues 5 and 10 on [this form](https://docs.google.com/spreadsheets/d/1s53eGMBj2hdTVLRitwi673R-_uERWh0kOxOp9QsOVg0/edit#gid=0).

Trello card: https://trello.com/c/vv0OR35g/630-march-update-tech-docs-template


lib/assets/stylesheets/modules/_search.scss
```diff
@@ -19,6 +12,48 @@
   }
 }
 
+.js .search__label {
+  @extend .govuk-visually-hidden;
+}
+
+.search__form {
+  position: relative;
+  width: 100%;
```
andysellick, line 21: Is the width needed here? This is on a form, which by default is a block level element - unless something else is overriding that that I can't see?

lib/assets/stylesheets/modules/_search.scss
```diff
@@ -19,6 +12,48 @@
...
+  bottom: 0;
+  right: 0;
+  border: 0;
+  background-color: govuk-colour('blue');
+  color: #fff;
+  cursor: pointer;
+  border-radius: 0;
+  padding: 0;
+  width: $input-size;
+  height: $input-size;
```
andysellick, line 40: With the height as a fixed px size if you zoom in on the page in Firefox with 'Zoom text only' turned on, the button isn't tall enough next to the input.

Screenshot 2021-03-08 at 13 10 19

This can be fixed by changing the height to `100%`.

lib/source/layouts/_search.erb
```diff
@@ -1,16 +1,22 @@
...
       Search (via Google)
     </label>
-    <input class="govuk-input govuk-!-margin-bottom-4" id="search" name="q" type="text" aria-controls="search-results" placeholder="Search">
+    <input
+      type="text"
+      id="search" name="q"
+      class="govuk-input govuk-!-margin-bottom-0 search__input"
+      aria-controls="search-results"
+      placeholder="Search">
+    <button type="submit" class="govuk-button search__button">Search</button>
```
andysellick, line 14: I'm not sure what the intended behaviour is for this button, but clicking on it without typing anything in first makes the search index panel appear like this:

Screenshot 2021-03-08 at 13 14 02

It should instead probably show the 'no results' state.

Screenshot 2021-03-08 at 13 15 40

lib/assets/stylesheets/modules/_search.scss
```diff
@@ -39,6 +74,7 @@ html.has-search-results-open {
     right: 0;
     bottom: 0;
     z-index: 600;
+    min-height: 100vh;
```
andysellick, line 77: On desktop, this line seems to be overridden by line 86. On mobile, turning it off doesn't seem to make any difference. Have I missed what this is supposed to do?

lib/assets/stylesheets/modules/_search.scss
```diff
@@ -68,6 +105,7 @@ html.has-search-results-open {
   }
 .search-results__inner {
   position: relative;
+  height: 100%;
```
andysellick, line 108: Can't see what this line does, can you explain?

andysellick (commented):
Have done a quick review, some questions. Happy to chat offline if I've misunderstood some things.

lib/source/layouts/_search.erb
danacotoran: Hmmmm. The behaviour when you click the button without typing anything in seems to be the same as the behaviour when you press Enter with the empty field in focus on the [current version of tech docs](https://docs.publishing.service.gov.uk/). This wasn't flagged as an issue in the report, that I could tell. 😕  But maybe it should have been?

lib/assets/stylesheets/modules/_search.scss
danacotoran: Thanks for flagging this, I forgot to test this with text magnification. Have implemented the solution you suggested.

lib/assets/stylesheets/modules/_search.scss
danacotoran: Nope, I don't think it should be needed.

lib/assets/stylesheets/modules/_search.scss
danacotoran: I think it's the same as [here](https://github.com/alphagov/tech-docs-gem/pull/209#discussion_r589415139). I've removed this line.

lib/source/layouts/_search.erb
danacotoran: Alright @andysellick  I think this should be fixed now

lib/assets/stylesheets/modules/_search.scss
```diff
@@ -19,6 +12,48 @@
...
+
+.search__button {
+  position: absolute;
+  bottom: 0;
+  right: 0;
+  border: 0;
+  background-color: govuk-colour('blue');
+  color: #fff;
+  cursor: pointer;
+  border-radius: 0;
```
andysellick, line 36: Is border-radius needed? Is that for a specific browser?

andysellick (approved):

lib/assets/stylesheets/modules/_search.scss
danacotoran: Nope, I cross-browser tested and it doesn't seem like it's needed after all. I think I had just copied it over from the GOVUK search button styles.

PR #214, opened 2021-03-08
Implement row level table headings to allow accessible tables with row headings

richardTowers:
Sometimes, tables have row-level headings as well as column level headings.

To be accessible, these need to be marked up as `<th>` tags. Otherwise users of screen readers and other assistive technology may not be able to work out that they're table headings rather than ordinary table cells.

⚠️ Don't forget to update the gem version in the [version file](https://github.com/alphagov/tech-docs-gem/blob/master/lib/govuk_tech_docs/version.rb) and [CHANGELOG](https://github.com/alphagov/tech-docs-gem/blob/master/CHANGELOG.md) before merging! ⚠️

### DIFF
--- a/lib/govuk_tech_docs/tech_docs_html_renderer.rb
+++ b/lib/govuk_tech_docs/tech_docs_html_renderer.rb
@@ -30,5 +30,34 @@ module GovukTechDocs
         </table>
       </div>)
     end
+
+    def table_row(body)
+      # Post-processing the table_cell HTML to implement row headings.
+      #
+      # Doing this in table_row instead of table_cell is a hack.
+      #
+      # Ideally, we'd use the table_cell callback like:
+      #
+      # def table_cell(content, alignment, header)
+      #   if header
+      #     "<th>#{content}</th>"
+      #   elsif content.start_with? "# "
+      #     "<th scope="row">#{content.sub(/^# /, "")}</th>"
+      #   else
+      #     "<td>#{content}</td>"
+      #   end
+      # end
+      #
+      # Sadly, Redcarpet's table_cell callback doesn't allow you to distinguish
+      # table cells and table headings until https://github.com/vmg/redcarpet/commit/27dfb2a738a23aadd286ac9e7ecd61c4545d29de
+      # (which is not yet released). This means we can't use the table_cell callback
+      # without breaking column headers, so we're having to hack it in table_row.
+      #
+      # Despite the hackiness, this substitution should work for all cases
+      # where the table cell just includes simple text (with no other markup).
+
+      body_with_row_headers = body.gsub(/<td># ([^<]*)<\/td>/, "<th scope=\"row\">\\1</th>")
+      "<tr>#{body_with_row_headers}</tr>"
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


36degrees:
Looking at https://github.com/vmg/redcarpet/commit/27dfb2a738a23aadd286ac9e7ecd61c4545d29de, I'm not sure it actually helps with table row headers – I think it just makes it possible to know whether a given cell is a _column header_ within the context of the `table_cell` method if you're writing your own renderer?

richardTowers:
Yeah, the problem is if I implement a custom `table_cell` callback, all the column headers come out as `<td>`s - there's not enough information in the callback to work out what's a header and what's a cell.

So I could use it to implement row headers, but we'd lose column headers (which work at the moment).

36degrees:
Ah! I totally mis-interpreted your comment. I was interpreting it as though that commit would bring native row headers to redcarpet, meaning we don't have to do anything at all – rather than that it would allow us to instead use the `table_cell` callback which would be less hacky.

Makes sense now 👍🏻

richardTowers:
Updated to clarify the comment, and to use `table_row` instead of `table`, which is a little bit more specific. Test still passes by the looks of it.

lib/govuk_tech_docs/tech_docs_html_renderer.rb
```diff
@@ -30,5 +30,34 @@ def table(header, body)
...
+      #
+      # Sadly, Redcarpet's table_cell callback doesn't allow you to distinguish
+      # table cells and table headings until https://github.com/vmg/redcarpet/commit/27dfb2a738a23aadd286ac9e7ecd61c4545d29de
+      # (which is not yet released). This means we can't use the table_cell callback
+      # without breaking column headers, so we're having to hack it in table_row.
+      #
+      # Despite the hackiness, this substitution should work for all cases
+      # where the table cell just includes simple text (with no other markup).
+
+      body_with_row_headers = body.gsub(/<td># ([^<]*)<\/td>/, "<th scope=\"row\">\\1</th>")
```
36degrees, line 59: Am I right in thinking the space between the `#` and the text is required for this pattern to match?

From testing, it looks like redcarpet will happily make something a heading without the space (e.g. `#Hello world` becomes an H1), so I'm wondering if it's worth being consistent on that here.

(Conversely, CommonMark [requires the space, but does acknowledge that many implementations don't](https://spec.commonmark.org/0.29/#atx-headings))

spec/govuk_tech_docs/tech_docs_html_renderer_spec.rb
```diff
@@ -0,0 +1,35 @@
+RSpec.describe GovukTechDocs::TechDocsHTMLRenderer do
+  describe "#render a table" do
+    it "renders tables with row headings" do
```
36degrees, line 3: Nice test 💯

36degrees (commented):
Makes sense to me. I have got 1 query about the space after the `#` being required.

I'm also aware it's been a long time since I've written any Ruby, so it might be worth getting a second review from someone.

lib/govuk_tech_docs/tech_docs_html_renderer.rb
richardTowers: Yeah, this pattern requires it. I wasn't sure whether it was needed in the spec or not.

🤔  🤔  Not sure whether to keep the space or not...

AlanGabbianelli (approved):
I like the use of Nokogiri here. ⭐
It's verbose, true, but it's nice and readable. It's clear what you're doing at each step.
As suggested in our Slack conversation, I'd like to see the test split but that's not a blocker 😉

AlanGabbianelli:
I see you split the test while I was still writing my comment 😄 Nice

PR #215, opened 2021-03-11
Add tabindex and focus states to code blocks

richardTowers:
This is required for accessibility reasons. Because some code blocks have scrollbars they need to be focusable using a keyboard.

I've copied the styles from the design system's documentation (see the code blocks in the HTML / Nunjucks tabs here: https://design-system.service.gov.uk/components/button/)

The implementation is simple when syntax highlighting is not enabled, because we can simply implement `block_code` to surround the code with the markup we want.

When syntax highlighting is enabled it's a bit more complicated.  Middleman already `include`s its own implementation of `block_code`, which renders the source code highlighted with spans. This method doesn't look like it would be easy to customise, so instead I've resorted to post-processing the HTML (i.e. replacing the attributes on the `<pre>` tag using a regex).

Testing this was a bit tricky, because you need to monkey patch the class to include the `block_code` method. I had to clone the class and patch the clone to do avoid affecting other tests.

Example from a locally running version of the paas tech docs:

image

-----

### DIFF
--- a/lib/assets/stylesheets/modules/_technical-documentation.scss
+++ b/lib/assets/stylesheets/modules/_technical-documentation.scss
@@ -123,6 +123,12 @@
     overflow: auto;
     position: relative;
     border: 1px solid $code-02;
+
+    &:focus {
+      padding: 13px;
+      border: $govuk-focus-width solid $govuk-focus-text-colour;
+      outline: $govuk-focus-width solid $govuk-focus-colour;
+    }
   }
 
   pre code {
--- a/lib/assets/stylesheets/modules/_technical-documentation.scss
+++ b/lib/assets/stylesheets/modules/_technical-documentation.scss
@@ -118,14 +118,17 @@
   }
 
   pre {
+    $padding: 15px;
+    $border-width: 1px;
+
     background: $code-00;
-    padding: 15px;
+    padding: $padding;
     overflow: auto;
     position: relative;
-    border: 1px solid $code-02;
+    border: $border-width solid $code-02;
 
     &:focus {
-      padding: 13px;
+      padding: $padding - ($govuk-focus-width - $border-width);
       border: $govuk-focus-width solid $govuk-focus-text-colour;
       outline: $govuk-focus-width solid $govuk-focus-colour;
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - [#210: Fix issue with WCAG 2.1 success criterion 1.3.1 (Info and Relationships)](https://github.com/alphagov/tech-docs-gem/pull/210)
 - [#209: Some search and keyboard navigation updates](https://github.com/alphagov/tech-docs-gem/pull/209)
 - [#214: Implement row level table headings to allow accessible tables with row headings](https://github.com/alphagov/tech-docs-gem/pull/214)
+- [#215: Add tabindex and focus states to code blocks](https://github.com/alphagov/tech-docs-gem/pull/215)
 
 ### Ruby version bump
 
--- a/lib/govuk_tech_docs/tech_docs_html_renderer.rb
+++ b/lib/govuk_tech_docs/tech_docs_html_renderer.rb
@@ -74,5 +74,23 @@ module GovukTechDocs
 
       tr.to_html
     end
+
+    def block_code(text, lang)
+      if defined?(super)
+        # Post-processing the block_code HTML to implement tabbable code blocks.
+        #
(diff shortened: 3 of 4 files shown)


lib/assets/stylesheets/modules/_technical-documentation.scss
```diff
@@ -125,6 +125,15 @@
     border: 1px solid $code-02;
   }
 
+  pre.code-block {
```
36degrees, line 128: Do we need the class here, or could we just blanket apply this to all `<pre>` elements within `.technical-documentation`?

Given any `pre` could be scrollable, it seems to make sense to tie the two together. Equally, if there are code blocks without a tabindex, the focus states won't apply to them anyway.

lib/assets/stylesheets/modules/_technical-documentation.scss
```diff
@@ -125,6 +125,15 @@
...
+    border: $govuk-focus-width solid transparent;
```
36degrees, line 129: This is overriding the border on `pre` (line 125), which means code blocks no longer have a border:

## Before

Screenshot 2021-03-11 at 15 18 30

## After

Screenshot 2021-03-11 at 15 18 24

We use the transparent border in the Design System so that the box 'shrink' when focused. This works there because we don't have a (visible) border on code blocks.

An alternative that doesn't change the design might be to change the padding when focused – reducing the padding from 15px to 13px to compensate for the extra 2px on the border.

lib/assets/stylesheets/modules/_technical-documentation.scss
richardTowers: Yeah, agreed. The class is not necessary.

Resolved in this force push: https://github.com/alphagov/tech-docs-gem/compare/67994d8b1e406b54e10587c57b228e359d1b5f22..0f5303b52c286d2aa3d5e51f65f9e651a5672fab

lib/assets/stylesheets/modules/_technical-documentation.scss
```diff
@@ -123,6 +123,12 @@
     overflow: auto;
     position: relative;
     border: 1px solid $code-02;
+
+    &:focus {
+      padding: 13px;
```
36degrees, line 128: I'd be tempted to add a comment explaining where this 'magic number' has come from / what it's doing.

Alternatively, you could 'show your working' by doing something like this:

```scss
  pre {
    $padding: 15px;
    $border-width: 1px;

    background: $code-00;
    padding: $padding;
    overflow: auto;
    position: relative;
    border: $border-width solid $code-02;

    &:focus {
      padding: $padding - ($govuk-focus-width - $border-width);
      border: $govuk-focus-width solid $govuk-focus-text-colour;
      outline: $govuk-focus-width solid $govuk-focus-colour;
    }
}
```

That also means everything's more likely to stay in sync if e.g. someone came in and changed the padding or border width.

lib/assets/stylesheets/modules/_technical-documentation.scss
richardTowers: Good spot, thanks! My backend developer eyes can't see things like missing borders 🙈 

I've implemented your suggestion of dropping two pixels of padding in this force push: https://github.com/alphagov/tech-docs-gem/compare/67994d8b1e406b54e10587c57b228e359d1b5f22..0f5303b52c286d2aa3d5e51f65f9e651a5672fab

image

image

lib/assets/stylesheets/modules/_technical-documentation.scss
richardTowers: Yep, good shout. Updated in 6467c8abc13497482da62f12ce4b16c6a3a1460e

36degrees (approved):

AlanGabbianelli (commented):
Looks good 👍
Just a note on the test organisation and semantic. It would be easier to follow with a couple of minor changes:
- only one `describe` block to describe what you're testing
- two `context` blocks to specify the conditions of the tests

It would look something like this:
```
  describe "#render a code block" do
  	context "without syntax highlighting" do
      # setup for conditions without syntax highlighting

      it "sets tab index to 0"

      it "renders the code without syntax highlighting"

    context "with syntax highlighting" do
      # setup for conditions with syntax highlighting

      it "sets tab index to 0"

      it "renders the code with syntax highlighting"
    end
  end
```

richardTowers:
Thanks! Addressed in a0c9bfddbc2749514a17c867ac3f3512a388542d

AlanGabbianelli (approved):

PR #216, opened 2021-03-12
Release 2.2.0

richardTowers:
None of the accessibility fixes are breaking changes (to the best of my knowledge). They're more than just bugfixes though, so this should be a semver minor release.

Fixes #211

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,18 @@
 # Changelog
 
-## Unreleased
+## 2.2.0
+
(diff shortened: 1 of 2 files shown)


richardTowers:
Converting to draft while I check if we want to bump some dependencies as part of this release.

AlanGabbianelli (approved):

richardTowers:
Nope, the dependency issues are all in downstream repos, so this is ready for release.

PR #223, opened 2021-03-23
ISSUE 220 - Remove unused CSS 'govuk-button' class from the 'search' submit button

iqbalgds:
- The 'search' submit button had the un-necessary CSS class 'govuk-button'.
- It a someone uses the gem and they are importing the Design System 'govuk-button' CSS stylesheet,
  then it breaks the styling on the button.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.2.2
+
+- [#223: Remove unnecessary CSS class on the search submit button](https://github.com/alphagov/tech-docs-gem/pull/223)
+
 ## 2.2.1
 
 - [#218: Remove unnecessary explicit dependencies: sprockets, activesupport, sass and pry](https://github.com/alphagov/tech-docs-gem/pull/218)
--- a/lib/source/layouts/_search.erb
+++ b/lib/source/layouts/_search.erb
@@ -11,7 +11,7 @@
       class="govuk-input govuk-!-margin-bottom-0 search__input"
       aria-controls="search-results"
       placeholder="Search">
-    <button type="submit" class="govuk-button search__button">Search</button>
+    <button type="submit" class="search__button">Search</button>
   </form>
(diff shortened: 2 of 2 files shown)


lib/govuk_tech_docs/version.rb
```diff
@@ -1,3 +1,3 @@
 module GovukTechDocs
-  VERSION = "2.2.1".freeze
+  VERSION = "2.2.2".freeze
```
36degrees, line 2: I think we tend to do releases separately in this repo, rather than making a change and releasing in the same PR.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.2.2
+
+- [#218: Remove unnecessary CSS class on the search submit button](https://github.com/alphagov/tech-docs-gem/pull/223)
```
36degrees, line 5:
```suggestion
- [#223: Remove unnecessary CSS class on the search submit button](https://github.com/alphagov/tech-docs-gem/pull/223)
```

36degrees (commented):
Thanks for picking this up @iqbalgds 🙌🏻

lib/govuk_tech_docs/version.rb
iqbalgds: Thanks - will update

CHANGELOG.md
iqbalgds: Good spot

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.2.2
```
36degrees, line 3: This should be 'Unreleased' until a new version is released (which might be 2.2.2, but could equally be 2.3) – it also makes it clear to users looking at the changelog that these changes haven't gone out yet.

richardTowers (approved):
Looks good to me (although I agree with @36degrees  about the title in the changelog, so you should probably fix that before merging 😄)

PR #224, opened 2021-03-24
ISSUE 221 - Tabbing on search results in mobile

iqbalgds:
- In mobile view, when you tab through the search results and tab past
  the last result:
  - The focus appears lost as it  goes to a 'close table of contents' button that
    is not in view.
- Update the SASS to hide the button when the search results are open.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 # Changelog
 
-## 2.2.2
+## Unreleased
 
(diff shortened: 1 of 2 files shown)


richardTowers (approved):
Looks good to me.

As with the other PR, the changelog heading should probably be "Unreleased", so you can bundle this up with any other changes you want to release.

PR #227, opened 2021-05-05
WIP: Add cookie banner

kr8n3r:
## What

Implement a configurable cookie banner based on the [GOV.UK Design system cookie banner component](https://design-system.service.gov.uk/components/cookie-banner/)

## How to test

- run locally
- experiment with configuration options ([example config](https://github.com/alphagov/tech-docs-gem/blob/6256243e2bb1ef19a688b344c1b42d85961b16b4/example/config/tech-docs.yml))

To do:
- [x] conditionally show different parts of the cookie banner based on config
- [x] [ document configuration](https://github.com/alphagov/tdt-documentation/pull/153)
- [ ] add tests

### DIFF
--- a/example/config/tech-docs.yml
+++ b/example/config/tech-docs.yml
@@ -52,3 +52,12 @@ api_path: source/pets.yml
 # Optional global settings for the page review process
 owner_slack_workspace: gds
 default_owner_slack: '#2nd-line'
+
+cookie_banner:
+  enable: true
+  cookie_types: ['essential','analytics', 'additional']
+  cookie_message_text: |
+    This is my custom text
+    This adds a line break.
+
+    This adds a new paragraph.
--- a/lib/source/layouts/_cookie_banner.erb
+++ b/lib/source/layouts/_cookie_banner.erb
@@ -1,0 +1,90 @@
+<%
+  is_enabled ||= config.tech_docs.dig(:cookie_banner, :enable)
+
+  essential_cookies = config.tech_docs.dig(:cookie_banner, :cookie_types) ? config[:tech_docs][:cookie_banner][:cookie_types].include?('essential') : false
+  analytics_cookies = config.tech_docs.dig(:cookie_banner, :cookie_types) ? config[:tech_docs][:cookie_banner][:cookie_types].include?('analytics') : false
+  additional_cookies = config.tech_docs.dig(:cookie_banner, :cookie_types) ? config[:tech_docs][:cookie_banner][:cookie_types].include?('additional') : false
+
+  custom_cookie_message_text = config.tech_docs.dig(:cookie_banner, :cookie_message_text)
+
+  cookie_page_path = config.tech_docs.dig(:cookie_banner, :cookie_page_path) ?
+    config[:tech_docs][:cookie_banner][:cookie_page_path] : '/cookies'
+
+  nonessential_cookie_type_text = additional_cookies ? 'additional' : 'analytics'
+%>
+
+<% cookie_message_text = capture do %>
+  <% if essential_cookies %>
+    <p>We use some essential cookies to make this service work.</p>
+  <% end %>
+  <% if additional_cookies %>
+    <p>We’d like to set additional cookies so we can remember your settings, understand how people use the service and make improvements.</p>
+  <% elsif analytics_cookies %>
+    <p>We’d also like to use analytics cookies so we can understand how you use the service and make improvements.</p>
+  <% end %>
+<% end %>
+
+<% if is_enabled %>
+<div class="govuk-cookie-banner " role="region" aria-label="Cookies on <%= config[:tech_docs][:service_name] %>">
+  <div class="govuk-cookie-banner__message govuk-width-container">
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <h2 class="govuk-cookie-banner__heading govuk-heading-m">Cookies on <%= config[:tech_docs][:service_name] %></h2>
+        <div class="govuk-cookie-banner__content">
+          <% if custom_cookie_message_text || cookie_message_text %>
+            <%= custom_cookie_message_text ? simple_format(custom_cookie_message_text) : cookie_message_text %>
+          <% end %>
+        </div>
+      </div>
+    </div>
+    <div class="govuk-button-group">
+      <% if analytics_cookies || additional_cookies %>
+        <button value="accept" type="button" name="cookies" class="govuk-button" data-module="govuk-button" data-accept-cookies="true">
+          Accept <%= nonessential_cookie_type_text %> cookies
+        </button>
+        <button value="reject" type="button" name="cookies" class="govuk-button" data-module="govuk-button" data-accept-cookies="false">
+          Reject <%= nonessential_cookie_type_text %> cookies
+        </button>
+      <% end %>
+      <% if essential_cookies %>
+        <a class="govuk-link" href="<%= cookie_page_path %>">View cookies</a>
+      <% end %>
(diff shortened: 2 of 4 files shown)


example/config/tech-docs.yml
```diff
@@ -52,3 +52,11 @@ api_path: source/pets.yml
 # Optional global settings for the page review process
 owner_slack_workspace: gds
 default_owner_slack: '#2nd-line'
+
+# cookie_banner: 
+#   enable: true
+#   cookie_types: ['essential','analytics', 'additional']
+#   essential_cookies_text: 'Essential cookies'
+#   analytics_cookies_text: 'Analytics cookies'
+#   additional_cookies_text: 'Additional cookies'
```
vanitabarrett, line 61: Would it be easier (and match more what the GOV.UK Design System component does) to have one variable for setting text rather than a differently named variable depending on the cookie types?

kr8n3r: easier: probably no different
match DS component logic: yes

vanitabarrett: It might help to simplify the code and might make it easier for users to understand how to override the message. E.g: if we need to show a custom message because we're implementing sub-domain consent, it becomes less obvious which string to override when we just need a whole new message.

kr8n3r: update to one variable and if present replaces default content.
there are still conditionals in places for when this text is not provided

lib/source/layouts/_cookie_banner.erb
```diff
@@ -0,0 +1,108 @@
...
+
+  additional_cookies = config.tech_docs.dig(:cookie_banner, :cookie_types) ? config[:tech_docs][:cookie_banner][:cookie_types].include?('additional') : false
+
+  custom_cookies_text = config.tech_docs.dig(:cookie_banner, :custom_cookies_text)
+
+  default_essential_cookies_text = "We use some essential cookies to make this service work."
+
+  default_analytics_cookies_text = "We’d also like to use analytics cookies so we can understand how you use the service and make improvements."
+
+  default_additional_cookies_text = "We’d like to set additional cookies so we can remember your settings, understand how people use the service and make improvements."
```
vanitabarrett, line 15: I have a feeling this could be simplified further... I had a go in a branch which you can see here: https://github.com/alphagov/tech-docs-gem/compare/add-cookie-banner...refactor-cookie-message

Let me know what you think!

lib/source/layouts/_cookie_banner.erb
```diff
@@ -0,0 +1,108 @@
...
+          <button value="accept" type="button" name="cookies" class="govuk-button" data-module="govuk-button">
+            Accept <%= nonessential_cookie_type_text %> cookies
+          </button>
+          <button value="reject" type="button" name="cookies" class="govuk-button" data-module="govuk-button">
+            Reject <%= nonessential_cookie_type_text %> cookies
+          </button>
+        <% end %>
+        <% if essential_cookies %>
+          <a class="govuk-link" href="<%= cookie_page_path %>">View cookies</a>
+        <% end %>
```
vanitabarrett, line 61: Can we include this all the time? I think we recommend always having a cookie page when a cookie banner is present, so I don't think it matters what type of cookies they're using?

lib/source/layouts/_cookie_banner.erb
```diff
@@ -0,0 +1,108 @@
...
+        <% if essential_cookies && !(analytics_cookies || additional_cookies) %>
+          <button type="button" class="govuk-button" data-module="govuk-button">
+            Hide this message
+          </button>
```
vanitabarrett, line 65: Do we need to add `data-` attributes or scoped classes for teams to hook into with their own JS? I'm thinking about teams needing to add the 'hide this message' functionality, and teams needing to show/hide cookie and confirmation messages.

vanitabarrett (changes requested):
Thanks for working on this, Jani. Just got a couple of comments, hope they make sense!

I also noticed when running this locally and looking at the Design System frontend docs site that URLs seem to link directly to the first heading on a page, which means the cookie banner is automatically scrolled out of view. I'm not sure where that behaviour comes from, but I wonder what the reason was behind it and if it's something teams could turn off in the config (otherwise it's unlikely anyone will see the banner) ?

lib/source/layouts/_cookie_banner.erb
kr8n3r: > Do we need to add `data-` attributes or scoped classes for teams to hook into with their own JS? I'm thinking about teams needing to add the 'hide this message' functionality, and teams needing to show/hide cookie and confirmation messages.

probably yes, we currently have `data-accept-cookies="true/false"` same as gov.uk so best just to add those for consistency

lib/source/layouts/_cookie_banner.erb
kr8n3r: I was thinking that could alway be there , but what if a service has no essential cookies only say analytics, but have no need for a custom text message? 
probably unlikely scenario tbh

lib/source/layouts/_cookie_banner.erb
kr8n3r: yeah makes sense and removes the need for extra variables as well

kr8n3r:
> Thanks for working on this, Jani. Just got a couple of comments, hope they make sense!
> 
> I also noticed when running this locally and looking at the Design System frontend docs site that URLs seem to link directly to the first heading on a page, which means the cookie banner is automatically scrolled out of view. I'm not sure where that behaviour comes from, but I wonder what the reason was behind it and if it's something teams could turn off in the config (otherwise it's unlikely anyone will see the banner) ?


it looks like it's driven by `in-page-navigation.js` I don't think there's a config setting for this

https://docs.cloud.service.gov.uk/ imports a markdown file for the first page and that somehow circumvents the JS

lib/source/layouts/_cookie_banner.erb
kr8n3r: updated, thanks

lib/source/layouts/_cookie_banner.erb
kr8n3r: added

vanitabarrett: @kr8n3r Can they be added to all the Hide message buttons? Reject banner button is missing one https://github.com/alphagov/tech-docs-gem/pull/227/files#diff-94a55eb7b630d0cb9c16329983ce87f649ca416e784207396d58506593859fbdR83

kr8n3r:
Build is failing.See https://github.com/alphagov/tech-docs-gem/issues/230 for detail

lib/source/layouts/_cookie_banner.erb
kr8n3r: done

m-green:
Hi @kr8n3r  - thanks for working on this! Was it ready for another review or did you need to do more work on it? (I've added @lfdebrux as a reviewer for when it's ready.)

@jonathanglassman Are you ok to work with @kr8n3r on [the related documentation PR](https://github.com/alphagov/tdt-documentation/pull/153)? I'm happy to pre-i or 2i it.

kr8n3r:
> Hi @kr8n3r - thanks for working on this! Was it ready for another review or did you need to do more work on it? (I've added @lfdebrux as a reviewer for when it's ready.)
> 
> @jonathanglassman Are you ok to work with @kr8n3r on [the related documentation PR](https://github.com/alphagov/tdt-documentation/pull/153)? I'm happy to pre-i or 2i it.

 it does require writing tests for it, but that would mean setting up a new test framework.
It worked on this when I had some time (which was a while ago) but it's not of any priority for us at the moment

PR #233, opened 2021-05-14
Fix failing builds on Travis

### LINKED ISSUE #230: Failing builds
due to https://github.com/ai/autoprefixer-rails/issues/203

project is using [middleman-autoprefixer](https://github.com/middleman/middleman-autoprefixer)

According to their [changelog](https://github.com/middleman/middleman-autoprefixer/blob/master/CHANGELOG.md), latest version is 3.0.0 which includes autoprefixer v10, but [rubygems](https://rubygems.org/gems/middleman-autoprefixer) only lists v2.10.1

lfdebrux:
Fixes #230.

This PR changes the Travis configuration to make sure it is using the same version of JavaScript as developers, and also changes the gemspec of `govuk_tech_docs` so that the newest version of [autoprefixer-rails](github.com/ai/autoprefixer-rails) is used.

Together these changes allow the Travis builds to work correctly. Bumping `autoprefixer-rails` is not enough, because otherwise Travis will use Node.js v8.12.0 and autoprefixer will complain that the version of Node is too old.


kr8n3r (approved):
:shipit:

Issue #222, opened 2021-03-23
Search has weird behaviour


lfdebrux:
### What I'm trying to do

Search for "reliability engineering"

### What I expect to happen

As I type the word "reliability" the number of results decreases steadily to only those pages containing the word "reliability"

### What actually happens

I type "reliab", I get 10 results.

I type "reliabi", or "reliabil", or "reliabili", or "reliabilit", I get 0 results.

I type "reliability", I get 10 results.

I've attached a screen recording of this happening.

https://user-images.githubusercontent.com/503614/112124302-1c4e3f00-8bba-11eb-9090-99ddbe236f1a.mp4

lfdebrux:
Thanks for transferring @36degrees 

It looks like search pipeline is configured in `govuk_tech_docs.rb`

lfdebrux:
`middleman-search` is based on LunrJS so the fix from Design System might still work, I am happy to investigate.

lfdebrux:
I can reproduce with the example in this repo using `documentatio` -> `documentation`

lfdebrux:
So fixing the behaviour is very straightforward; I am just trying to work it how to write tests for it!

PR #235, opened 2021-06-02
Improve search tests and fix formatting of search results

lfdebrux:
While working on a fix for issue #222 I found that I needed to update the dummy search index in the tests. I thought that it would be good to improve the search tests so that they instead use the `search.json` file from building the example site in this repo.

This PR makes that change by refactoring the `rebuild_site!` RSpec helper out of the RSpec configuration, and adding it to the `jasmine-gem` helpers as well. This means that before the JavaScript tests are run, the example site will be built. We can then replace the dummy search index with a AJAX request to the existing JSON file.

### Change to search results format

After doing this change I found that the tests were failing, because the dummy index was out-of-sync with the index produced by the production code. So this PR also changes the formatting of page HTML slightly, to add whitespace after heading tags, so that search results no longer have word collisions. This also means that headings in search results will no longer be followed by the first sentence of the section.

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<blockquote>
<h3>This is an expired page</h3>
<p>… This is an <strong>expired</strong> pageSee the banner on this page …</p>
</blockquote>
</td>
<td>
<blockquote>
<h3>This is an expired page</h3>
<p>… This is an <strong>expired</strong> page …</p>
</blockquote>
</td>
</tr>
<tr>
<td>

</td>
<td>

</td>
</tr>
</tbody>
</table>

I feel that this change should still be made though, as having word collisions is pretty bad.


spec/site_helper.rb
```diff
@@ -0,0 +1,31 @@
...
+module_function
+
+  # Build the example site using middleman
+
+  def rebuild_site!(config: "config/tech-docs.yml", overrides: {})
+    config_file = Tempfile.new("tech_docs_config")
+
+    begin
+      Dir.chdir("example") do
+        puts Dir.getwd
```
36degrees, line 14: Is this still needed? Or was this just for debugging?

36degrees (commented):
Looks great – and super commit messages throughout. Just one minor thing I spotted which I think is left over from debugging.

spec/site_helper.rb
lfdebrux: Oh nice spot, thanks!

36degrees (approved):

PR #238, opened 2021-06-21
Move the `aria-expanded` attribute to the correct element

colinbm:

## What

* Move the `aria-expanded` attribute onto the toggle button

## Why

From the DAC review:

> There are sections of content with Show/ Hide functionality. Neither the expandable behaviour of the control nor its expanded or collapsed state are programmatically determinable as the aria-expanded attribute has not been utilised (please note, it appears this attribute has been incorrectly associated with the expandable region rather than the trigger).

### DIFF
--- a/lib/assets/javascripts/_modules/collapsible-navigation.js
+++ b/lib/assets/javascripts/_modules/collapsible-navigation.js
@@ -34,10 +34,9 @@
           arrayOfIds.push(uniqueId)
           $(this).addClass('collapsible__body')
             .attr('id', uniqueId)
-            .attr('aria-expanded', 'false')
         })
         $heading.addClass('collapsible__heading')
-          .after('<button class="collapsible__toggle" aria-controls="' + arrayOfIds.join(' ') + '"><span class="collapsible__toggle-label">Expand ' + $heading.text() + '</span><span class="collapsible__toggle-icon" aria-hidden="true"></button>')
+          .after('<button class="collapsible__toggle" aria-expanded="false" aria-controls="' + arrayOfIds.join(' ') + '"><span class="collapsible__toggle-label">Expand ' + $heading.text() + '</span><span class="collapsible__toggle-icon" aria-hidden="true"></button>')
         $topLevelItem.on('click', '.collapsible__toggle', function (e) {
           e.preventDefault()
           var $parent = $(this).parent()
--- a/lib/assets/javascripts/_modules/collapsible-navigation.js
+++ b/lib/assets/javascripts/_modules/collapsible-navigation.js
@@ -47,14 +47,15 @@
     }
 
     function toggleHeading ($topLevelItem) {
-      var isOpen = $topLevelItem.hasClass('is-open')
+      var setOpen = !$topLevelItem.hasClass('is-open')
+
       var $heading = $topLevelItem.find('> a')
       var $button = $topLevelItem.find('.collapsible__toggle')
       var $toggleLabel = $topLevelItem.find('.collapsible__toggle-label')
 
-      $topLevelItem.toggleClass('is-open', !isOpen)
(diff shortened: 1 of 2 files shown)


36degrees:
> Whilst looking at this, I spotted an issue with the nav list that I've also fixed... currently we have too many `ul` elements. We should only have these at the open/close of the nav, and at the open/close of each new level of nav. As it is we have a `ul` element for _each_ `li`.

I'm not sure this change is quite right – now everything in the nav sits at the same level.

On the current version in master, the nav looks like this:

Screenshot 2021-06-22 at 09 41 53

On this branch, it looks like this:

Screenshot 2021-06-22 at 09 42 35

colinbm:
Odd - it's not doing that on mine - investigating.

lib/assets/javascripts/_modules/collapsible-navigation.js
```diff
@@ -49,11 +49,11 @@
...
       var isOpen = $topLevelItem.hasClass('is-open')
       var $heading = $topLevelItem.find('> a')
-      var $body = $topLevelItem.find('.collapsible__body')
+      var $button = $topLevelItem.find('.collapsible__toggle')
       var $toggleLabel = $topLevelItem.find('.collapsible__toggle-label')
 
       $topLevelItem.toggleClass('is-open', !isOpen)
-      $body.attr('aria-expanded', isOpen ? 'false' : 'true')
+      $button.attr('aria-expanded', isOpen ? 'false' : 'true')
       $toggleLabel.text(isOpen ? 'Expand ' + $heading.text() : 'Collapse ' + $heading.text())
```
36degrees, line 57: I appreciate this isn't code that you've written, but I found it quite confusing when reviewing and I'm wondering if there's anything we can do to make it clearer.

I thought that the `aria-expanded` attribute would be incorrect, because it's being set to `'false'` when `isOpen` is `true`, which seems backwards. However, `isOpen` is confusingly named, as it represents the state of the heading _before_ toggling, rather than the state after – which is what we care about when setting the `aria-expanded` attribute and the toggle text.

Any thoughts? Perhaps we could flip the boolean when assigning it at the top of the function rather than when calling `toggleClass`? Or rename the variable to make it clearer?

colinbm: Makes sense - refactored and updated.

36degrees:
Can we consider splitting the list structure changes into a separate PR? As I understand it, they're unrelated to the `aria-expanded` change which looks good to go?

colinbm:
@36degrees seperated that off

36degrees:
Thanks @colinbm! Would you mind tidying up the PR description to remove the changes that have been split out?

Unfortunately, there's another place in the code where `aria-expanded` is being set which needs fixing too. When the code is first initialised, `aria-expanded` is still getting set to `false` on the `collapsible__body`:

https://github.com/alphagov/tech-docs-gem/blob/4ba92c82317d59051bef93810089bc316af7e12f/lib/assets/javascripts/_modules/collapsible-navigation.js#L35-L37

This means that the collapsed state is not announced when you first navigate to a button, only once it's been toggled at least once.

We're also still including a non-sensical `aria-expanded` on the toggled region, except now it's permanently set to `false`.

We should really have tests that cover all of this included in [/spec/javascripts/collapsible-navigation-spec.js](https://github.com/alphagov/tech-docs-gem/blob/4ba92c82317d59051bef93810089bc316af7e12f/spec/javascripts/collapsible-navigation-spec.js). However, I don't think it's fair to ask you add them as you're only fixing the existing implementation.

colinbm:
@36degrees Ah, apologies - I had fixed that, but somehow it failed to go into the commit.

36degrees (approved):
👍🏻

PR #239, opened 2021-06-23
Add hints to search results to indicate it should be closed

colinbm:

## What

* Add `aria-hidden` attribute to main content (true when menu open).
* On pressing enter, focus on close link, rather than first result.
* Set `aria-modal` on search results, as hint that this is modal content and should be closed to return to main content.

## Why

DAC report

>
>

DAC proposed solution

> It may assist users to apply the WAI-ARIA design pattern for modal dialogs, including limiting keyboard focus and screen reading software in reading mode to the contents of the dialog when it is open. For more information, please refer to the [WAI-ARIA Authoring Practices 1.1 Modal Dialog Example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html). Otherwise, ensure that focus is taken to the ‘Close’ button upon activation.
>
> Please note, that there are currently varying degrees of support for the aria-modal attribute so it will not always trap screen reader focus in all browsers. It is also very important that when the modal is closed, the element with aria-modal="true" set is hidden with ‘display:none;’, otherwise it can prevent screen reader users from interacting with the page.

### DIFF
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -17,7 +17,6 @@
     var $searchResultsClose
     var results
     var query
-    var queryTimer
     var maxSearchEntries = 20
 
     this.start = function start ($element) {
@@ -51,25 +50,21 @@
     }
 
     function attach () {
-      // Search functionality on search text input
-      $searchInput.on('input', function (e) {
+      // Set focus on the results close link instead of submiting the search
+      // form to Google
+      $searchForm.on('submit', function (e) {
         e.preventDefault()
-        query = $(this).val()
+
+        query = $searchInput.val()
         s.search(query, function (r) {
           results = r
           renderResults(query)
           updateTitle()
         })
         if (window.ga) {
-          window.clearTimeout(queryTimer)
-          queryTimer = window.setTimeout(sendQueryToAnalytics, 1000)
+          sendQueryToAnalytics()
         }
-      })
 
-      // Set focus on the results close link instead of submiting the search
-      // form to Google
-      $searchForm.on('submit', function (e) {
-        e.preventDefault()
         showResults()
         var $closeButton = $searchResultsWrapper.find('.search-results__close')
         $closeButton.focus()
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -189,7 +189,7 @@
     // Default text is to display the number of search results
     function updateTitle (text) {
       if (typeof text === 'undefined') {
-        var count = results.length
+        var count = results ? results.length : 0
         var resultsText = count + ' results'
       }
       $searchResultsTitle.text(text || resultsText)
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -7,6 +7,7 @@
     var s = this
     var $html = $('html')
     var $tocNav
+    var $mainContent
     var $searchForm
     var $searchLabel
     var $searchInput
@@ -27,6 +28,7 @@
       $searchResults = $searchResultsWrapper.find('.search-results__content')
       $searchResultsTitle = $searchResultsWrapper.find('.search-results__title')
       $searchResultsClose = $searchResultsWrapper.find('.search-results__close')
+      $mainContent = $html.find('.app-pane__content')
       $tocNav = $('#toc')
       s.downloadSearchIndex()
       attach()
@@ -64,19 +66,14 @@
         }
       })
 
-      // Set focus on the first search result instead of submiting the search
+      // Set focus on the results close link instead of submiting the search
       // form to Google
       $searchForm.on('submit', function (e) {
         e.preventDefault()
         showResults()
-        var $firstResult = $searchResults.find('.search-result__title a').first()
-        if ($firstResult.length) {
-          $firstResult.focus()
-        } else {
-          // if there are no results, show the "0 results" state
-          results = []
-          updateTitle()
-        }
+        var $closeButton = $searchResultsWrapper.find('.search-results__close')
+        $closeButton.focus()
+        updateTitle()
       })
 
       // Closing the search results, move focus back to the search input
@@ -200,6 +197,7 @@
     function showResults () {
       $searchResultsWrapper.addClass('is-open')
         .attr('aria-hidden', 'false')
+      $mainContent.attr('aria-hidden', 'true')
       $html.addClass('has-search-results-open')
       $tocNav.addClass('search-results-open')
     }
@@ -207,6 +205,7 @@
     function hideResults () {
       $searchResultsWrapper.removeClass('is-open')
         .attr('aria-hidden', 'true')
+      $mainContent.attr('aria-hidden', 'false')
       $html.removeClass('has-search-results-open')
       $tocNav.removeClass('search-results-open')
     }
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -6,8 +6,11 @@
   Modules.Search = function Search () {
     var s = this
     var $html = $('html')
+    var $header
+    var $headerContent
     var $tocNav
     var $mainContent
+    var $searchContainer
     var $searchForm
     var $searchLabel
     var $searchInput
@@ -20,6 +23,9 @@
     var maxSearchEntries = 20
 
     this.start = function start ($element) {
+      $header = $html.find('.app-pane__header')
+      $headerContent = $header.find('.govuk-header__content')
+      $searchContainer = $html.find('.search')
       $searchForm = $element.find('form')
       $searchInput = $element.find('#search')
       $searchLabel = $element.find('.search__label')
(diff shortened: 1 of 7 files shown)


36degrees:
Thanks for looking at this. This is a really difficult interaction to make accessible, and unfortunately I don't think this is going to be quite as easy as DAC have made it sound!

> Add aria-hidden attribute to main content (true when menu open).

I don't believe this is doing anything, because the use of `visibility: hidden` on `technical-documentation` when the `has-search-results-open` class is applied to the `<html>` element means that the main content is already hidden from _everyone_, including screen readers.

[Best practise would be to not use `aria-hidden` in such cases](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute#best_practices).

> Set aria-modal on search results, as hint that this is modal content and should be closed to return to main content.

The use of [`aria-modal`](https://www.w3.org/TR/wai-aria-1.1/#aria-modal) on the search results pane does much more than that – it also means that **everything outside of the search results is hidden from screen readers**. However, we're not doing anything to ensure that the focus is retained within the search results pane. It's therefore possible to tab to elements outside of that area, and because those elements are hidden nothing gets announced. From the perspective of someone who cannot see the screen, they now have no idea where the focus is.

In this video, note how when the search results are displayed, the screen reader makes no announcement when the focus moves to the links in the navigation:

https://user-images.githubusercontent.com/121939/124277018-7c02a180-db3c-11eb-9ea1-50a87c77c728.mov

In terms of the WAI-ARIA modal dialog example, the missing piece of the puzzle is [the section on keyboard support](https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html#kbd_label), which dictates that when the focus is moved from the last element (or first element, when tabbing backwards) it should 'wrap' within the modal.

This is what DAC are referring to when they talk about 'focus trapping', and we're not currently doing it, which is what is breaking the pattern. I'd suggest spending some time looking at the example implementation to understand how this is meant to work.

Having said all of that, I'm not entirely sure that it does make sense to treat the search results page as a modal dialog. When the search results are visible the navigation remains visible and interactive if you're using a mouse or touch, so I'm not sure it makes sense to capture focus and prevent users being able to reach them using the keyboard.

And, even if you trap keyboard focus, a visually-sighted screen reader user may try to interact with the visible parts of the page (again, using the mouse or touch), but will not get any screen reader feedback.

To be honest, I'd suggest booking a slot at the next accessibility clinic to get input from an accessibility specialist. There's a lot of things to consider here…

colinbm:
I've gone back to DAC to enquire on this... it's possible that focussing on "Close" when submitted may be sufficient, but that doesn't really resolve the issue that the content disappears when typing.

They may come back and suggest that we remove the results as you type, and only display them on submit.

colinbm:
From DAC:

> I would say that the biggest problem is that you are hiding the page from assistive technology when a use types in the search field. Without the visual prompt of the form covering the content, there is no reason why a screen reader users would know that the dialog needs to be closed for the page to be read. There is no association between the close button they locate under the search field and the rest of the page content. I would strongly suggest allowing a user to explicitly submit the search rather than the search being submitted ‘live’ as a user types. I would then make the results dialog modal and would not allow users to access the left navigation when it was open. The only other alternative I can think of is to maintain the current implementation but not to hide the page when the results dialog pops up; however, this may be frustrating for keyboard only users, as the dialog covers half the page, so users would be focusing components which are visually hidden.
> 
> Ideally, I would really recommend an entirely separate search results page.

So as I thought, the *minimal* fix would be to remove the "live" updating of results. I'm still not sure about the modal and preventing nav access, unless we add a visual indication that it's disabled.

I think a separate page for search results would be a bigger job than planned for here so I won't look at that too much unless we decide on that directiom.

colinbm:
From DAC:

> Removing the live updating will certainly help, but I also think the main page content should not be hidden when the search dialog is open. A good way to visualise what a screen reader user would encounter is to disable the CSS and disregard everything we know will be hidden from assistive technology with display:none;.
>
> Sighted users can see the search results visually covering the main page content, but this is not so for a screen reader user – they encounter the content linearly in the order it appears in the DOM. The ‘Close search results’ button may be easily missed and there is no reason for a user to make the association between search results which appear in the navigation region and the main content disappearing much further down the page. Unless you can immediately see that making a search has triggered the main page content to be hidden, it would be very difficult to infer that connection.

Based on that, the latest change removes the live updating, leaves the content visible, and reduces the content's opacity when the search results are visible.

lib/assets/javascripts/_modules/search.js
```diff
@@ -49,34 +56,25 @@
...
-        var $firstResult = $searchResults.find('.search-result__title a').first()
-        if ($firstResult.length) {
-          $firstResult.focus()
-        } else {
-          // if there are no results, show the "0 results" state
-          results = []
-          updateTitle()
-        }
+        var $closeButton = $searchResultsWrapper.find('.search-results__close')
+        $closeButton.focus()
```
36degrees, line 76: I think we've already got a variable that references this element we can use here?

```suggestion
        $searchResultsClose.focus()
```

lib/assets/javascripts/_modules/search.js
```diff
@@ -200,13 +198,23 @@
     function showResults () {
       $searchResultsWrapper.addClass('is-open')
         .attr('aria-hidden', 'false')
+      $searchContainer.attr('aria-modal', 'true')
+      $mainContent.prop('inert', true)
+      $tocNav.prop('inert', true)
+      $header.prop('inert', true)
+      $headerContent.prop('inert', true)
```
36degrees, line 205: Given `$headerContent` is inside `$header`, and the `inert` property [affects all descendants](https://whatpr.org/html/4288/interaction.html#the-inert-attribute), I don't think we need to apply the attribute to both?

```suggestion
```

package-lock.json
```diff
@@ -1,7 +1,2505 @@
 {
   "name": "tech-docs-gem",
+  "lockfileVersion": 2,
```
36degrees, line 3: This is currently part of 469f2d492b09d4383dbee905e1eb819e865a7508 but has nothing to do with the rest of the commit. I think this is just because you're running npm v7 locally and so running `npm install` generates a new v2 lockfile? Suggest this change is made in a separate PR.

36degrees (commented):
This is looking good to me – just a couple of minor comments.

However, I think the PR title and description do not really reflect the change that's being made any more. Can we update them?

I think there's quite a lot of context in the correspondence with DAC that's worth capturing in the PR and/or commit messages too, so we have that history when we try to understand why we've made these changes in the future?

36degrees:
The tests are also failing:

> ReferenceError: Can't find variable: Map

Unfortunately, I think this is coming from the Inert polyfill we're now including.

package-lock.json
colinbm: Think this was related to installing inert (first tried through npm) - reverted that.

36degrees:
The approach outlined in this PR makes use of the `inert` attribute which is relatively new and still has patchy support in browsers. 

We're vendoring in a polyfill for it ([WICG/inert](https://github.com/WICG/inert), included as part of 4eff8ac461af154d8624524f6dc3f5a9241247f7) but that polyfill itself relies on [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set), [Element.prototype.matches](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches) and [Node.prototype.contains](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains) (as per https://github.com/WICG/inert#legacy-browser-support).

Of those:

- `Element.prototype.matches` works in IE9 and above – although it has a non-standard name, [the `inert` polyfill does cater for this](Element.prototype.matches)
- `Node.prototype.contains` works in IE9 and above on `HTMLElement` nodes only
- `Map` and `Set` are only supported in IE11 and above

Notably Jasmine, which we use to run our JS tests, [uses PhantomJS under the hood which does not support `Map` or `Set` either](https://github.com/searls/jasmine-maven-plugin/issues/269#issuecomment-299746490). **This is why the tests are currently failing with `ReferenceError: Can't find variable: Map`.**

Options at this point:

- re-consider an alternative approach that doesn't rely on the `inert` attribute (such as the approach explored in https://github.com/alphagov/tech-docs-gem/tree/search-modal-close-focus-lost)
- include a polyfill for Set and Map as part of the application JS – which would add extra weight to the page but would allow search to work in IE9 and above
- include a polyfill for Set and Map _when running the tests only_, – which would get the tests passing but search will fail non-gracefully in IE8-10 and any other browser that does not support Set / Map
- update Jasmine to use a different runner rather than phantomjs – which would get the tests passing, but search will fail non-gracefully in IE8/9/10 and any other browser that does not support Set / Map

36degrees:
Based on the size of the response from polyfill.io, I think polyfilling Map and Set would add about 34kb (roughly 8kb gzipped) to the application JS.

PR #240, opened 2021-06-24
Update menu html structure so it's one single hierarchical list

colinbm:

## What

Updates the navigation to use a single top level `ul` and nested subnav.

## Why 

This is a follow up to https://github.com/alphagov/tech-docs-gem/pull/237 with a better solution.

The previous markup was like this:

```html
<ul>
  <li>
    <a href="">Top level link</a>
  </li>
</ul>

<ul>
  <li>
    <a href="">Top level link</a>
    <ul>
      <li>indented nav</li>
    </ul>
  </li>
</ul>
```

This separates each top level nav item into a separate list, where the nav should be a single component.

New markup is:

```html
<ul>
  <li>
    <a href="">Top level link</a>
  </li>
  <li>
    <a href="">Top level link</a>
    <ul>
      <li>indented nav</li>
    </ul>
  </li>
</ul>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - [Bump redcarpet to 3.5.1 to fix CVE-2020-26298](https://github.com/alphagov/tech-docs-gem/pull/226)
+- [#240: Update menu html structure so it's one single hierarchical list](https://github.com/alphagov/tech-docs-gem/pull/240)
 
 ## 2.3.0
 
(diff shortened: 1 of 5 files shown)


36degrees:
I don't think b35334d93bebdd65f34d8d5b3e3d6fca46cdc445 belongs in this PR, but otherwise this seems to make sense to me.

colinbm:
@36degrees It's more adjacent... would you prefer it be in a separate PR? Or I can edit the title to include it explicitly?

36degrees:
I think it's more or less the same commit as https://github.com/alphagov/tech-docs-gem/pull/238/commits/f46aee7312d58e606cf81ac96018a1774fa73166 in #238? At the minute it'll conflict with that PR because it's changing the same lines of code but in slightly different ways.

colinbm:
Ah so it is - thanks for clarifying - will remove it here.

36degrees (approved):
Thanks @colinbm 👍🏻

PR #244, opened 2021-07-09
Don't change the focus of the page on initial load

colinbm:

## What

On loading a page initially `foo.html` do not change the URL to `foo.html#h1-id` until scrolling past.
Remove the fragment from the navigation for the `h1`.

## Why

It is expected that focus will begin at the start of the page. Changing this is confusing for screen reader users.

> On page load the main page heading is automatically focused. This may be navigationally disorientating for users reliant on the use of the keyboard alone to navigate.
> 
> Keyboard only user comments:
> “When this page loads, focus starts in the content of the page. I found this confusing as I expected my focus to start at the top of the page.”
> 
> Solution:
> Users would expect to navigate the page in reading order, from top to bottom and left to right.
>  

## Note

~I'm not sure about the navigation change... in itself it's not confusing as the target fragment is clear in the clicked link, but it still seems like the link to the "page" rather than "position on page" should be at the beginning. I'm going to double check this with DAC, but welcome any thoughts.~ Confirmed this is correct approach with DAC.

### DIFF
--- a/lib/assets/javascripts/_modules/in-page-navigation.js
+++ b/lib/assets/javascripts/_modules/in-page-navigation.js
@@ -47,15 +47,15 @@
         fragment = fragmentForFirstElementInView()
       }
 
-      handleChangeInActiveItem(fragment)
+      handleChangeInActiveItem(fragment, true)
     }
 
     function handleScrollEvent () {
       handleChangeInActiveItem(fragmentForFirstElementInView())
     }
 
-    function handleChangeInActiveItem (fragment) {
-      storeCurrentPositionInHistoryApi(fragment)
+    function handleChangeInActiveItem (fragment, skipHistory) {
+      !skipHistory && storeCurrentPositionInHistoryApi(fragment)
       highlightActiveItemInToc(fragment)
     }
 
--- a/lib/assets/javascripts/_modules/in-page-navigation.js
+++ b/lib/assets/javascripts/_modules/in-page-navigation.js
@@ -47,15 +47,13 @@
         fragment = fragmentForFirstElementInView()
       }
 
-      handleChangeInActiveItem(fragment, true)
+      highlightActiveItemInToc(fragment)
     }
 
     function handleScrollEvent () {
-      handleChangeInActiveItem(fragmentForFirstElementInView())
-    }
+      var fragment = fragmentForFirstElementInView()
 
-    function handleChangeInActiveItem (fragment, skipHistory) {
-      !skipHistory && storeCurrentPositionInHistoryApi(fragment)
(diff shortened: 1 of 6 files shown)


lib/assets/javascripts/_modules/in-page-navigation.js
```diff
@@ -47,15 +47,15 @@
         fragment = fragmentForFirstElementInView()
       }
 
-      handleChangeInActiveItem(fragment)
+      handleChangeInActiveItem(fragment, true)
     }
 
     function handleScrollEvent () {
       handleChangeInActiveItem(fragmentForFirstElementInView())
     }
```
36degrees, line 55: Given that `handleChangeInActiveItem` now only does 1 thing when `skipHistory` is true, what do you think about re-factoring this code to remove `handleChangeInActiveItem` entirely? I think this might make the intent clearer and avoids the need to introduce a new boolean flag.

1. Within `handleInitialLoadEvent`, call `highlightActiveItemInToc` directly.
2. Within `handleScrollEvent`, call both `storeCurrentPositionInHistoryApi` and `highlightActiveItemInToc`, passing the fragment for the first element in view.
3. Remove the `handleChangeInActiveItem` function (not included in suggestion)

```suggestion
      highlightActiveItemInToc(fragment)
    }

    function handleScrollEvent () {
      var fragment = fragmentForFirstElementInView()
      storeCurrentPositionInHistoryApi(fragment)
      highlightActiveItemInToc(fragment)
    }
```

What do you reckon?

lib/govuk_tech_docs/table_of_contents/heading.rb
```diff
@@ -12,8 +12,12 @@ def size
         @element_name.scan(/h(\d)/) && $1 && Integer($1)
       end
 
-      def href
-        @page_url + "#" + @attributes["id"]
+      def href(level)
```
36degrees, line 15: In theory, will level map to the type of heading? For example, will a level 1 always be an h1 and vice-versa? Just wondering if, rather than passing level in as an argument, we can use the existing `size` method.

lib/assets/javascripts/_modules/in-page-navigation.js
colinbm: 👍🏻  That makes sense!

lib/govuk_tech_docs/table_of_contents/heading.rb
colinbm: Ah yes - that'll be cleaner.

36degrees (approved):
👍🏻

Issue #254, opened 2021-07-29
FFI gem can cause issues for users on Mohave


tombye:
The `ffi` gem produces the following error for some users on Mohave (OSX 10.14.6):

```shell
dyld: lazy symbol binding failed: Symbol not found: _ffi_prep_closure_loc
  Referenced from: /Users/tombyers/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/ffi-1.15.3/lib/ffi_c.bundle
  Expected in: /usr/lib/libffi.dylib

dyld: Symbol not found: _ffi_prep_closure_loc
  Referenced from: /Users/tombyers/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/ffi-1.15.3/lib/ffi_c.bundle
  Expected in: /usr/lib/libffi.dylib
```

See https://github.com/ffi/ffi/issues/791 for more details.

I have personally seen in using the GDS way and Notify's tech docs. It has also been [an issue on Datagovuk's tech docs](https://github.com/alphagov/datagovuk-tech-docs/commit/d992c23c1d87d430a2e88b98709bc5fa85e710ed).

[@timblair suggested it would make sense to pin it in here](https://github.com/alphagov/gds-way/pull/632#issuecomment-889256345) instead of the downstream projects.

I'm happy to raise a PR to pin it if this seems sensible. Alternatively, I could add a note to the docs if that would be better.

m-green:
The tech writing team has also been experiencing this issue for a while. @PippaClarkGDS I seem to remember the Digital Identity team had a fix, but was it a workaround?

richardTowers:
For convenience, the options to workaround this from ffi/ffi#791 :

* Upgrade to macOS 10.15 (Catalina) or higher
* Tell rubygems not to use the system ffi (`gem inst ffi -- --disable-system-libffi`)
* Pin the ffi version back to 1.12.2 in the Gemfile

Also - if anyone's wondering what `ffi` is, it's the "Foreign Function Interface" library, that allows calls from ruby to native code (usually C). Why do we need it? It's a dependency of `sassc` (which is the C library we use to compile Sass) and `rb-inotify` (which middleman uses to detect file changes on linux when it's running in `serve` mode).

---

[...]

PR #255, opened 2021-07-30
Add note on ffi issue to README

tombye:
Users of apps consuming this gem and running on OSX Mohave (10.14) can experience an issue with the ffi gem.

This adds a note to the README listing fixes for the issue.

See https://github.com/alphagov/tech-docs-gem/issues/254 for related discussion.

I can see the warning about bumping the gem version but, looking at the history, this seems to be done as a separate PR, following those with code changes in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ gem 'govuk_tech_docs', path: '../tech-docs-gem'
 
 To preview your documentation changes locally, see the [Tech Docs Template documentation on previewing your documentation](https://tdt-documentation.london.cloudapps.digital/create_project/preview/#preview-your-documentation).
 
-If you experience [this issue](https://github.com/alphagov/tech-docs-gem/issues/254), possible fixes can be found [here](#issue-with-ffi-on-osx-mohave).
+If you experience [this issue with ffi](https://github.com/alphagov/tech-docs-gem/issues/254), possible fixes can be found [here](#issue-with-ffi-on-osx-mohave).
 
 ### Use the example in this repo
 
@@ -48,7 +48,7 @@ bundle exec middleman server
 
 See your website on `http://localhost:4567` in your browser.
 
-If you experience [this issue](https://github.com/alphagov/tech-docs-gem/issues/254), possible fixes can be found [here](#issue-with-ffi-on-osx-mohave).
+If you experience [this issue with ffi](https://github.com/alphagov/tech-docs-gem/issues/254), possible fixes can be found [here](#issue-with-ffi-on-osx-mohave).
 
 For more information on previewing your documentation locally, see the [Tech Docs template documentation on previewing your documentation](https://tdt-documentation.london.cloudapps.digital/create_project/preview/#preview-your-documentation).
 
@@ -67,9 +67,9 @@ To run the tests and see the results in your terminal, run:
 bundle exec rake jasmine:ci
 ```
 
-## Issue with FFI on OSX Mohave
+## Issue with FFI on OSX Mojave
 
-Users on OSX Mohave (10.14) may get this error when running `bundle exec middleman serve` on apps that use this gem.
+Users on OSX Mojave (10.14) may get this error when running `bundle exec middleman serve` on apps that use this gem.
 
 There are 3 different ways to solve this, depending on what is possible for you. Listed here as best to worst:
 
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ gem 'govuk_tech_docs', path: '../tech-docs-gem'
 
 To preview your documentation changes locally, see the [Tech Docs Template documentation on previewing your documentation](https://tdt-documentation.london.cloudapps.digital/create_project/preview/#preview-your-documentation).
 
-If you experience [this issue with ffi](https://github.com/alphagov/tech-docs-gem/issues/254), possible fixes can be found [here](#issue-with-ffi-on-osx-mohave).
+If you experience [the FFI gem issue for Mojave users](https://github.com/alphagov/tech-docs-gem/issues/254), you should refer to this [list of possible fixes](#issue-with-ffi-on-osx-mohave).
 
 ### Use the example in this repo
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,6 +34,8 @@ gem 'govuk_tech_docs', path: '../tech-docs-gem'
 
 To preview your documentation changes locally, see the [Tech Docs Template documentation on previewing your documentation](https://tdt-documentation.london.cloudapps.digital/create_project/preview/#preview-your-documentation).
 
+If you experience [this issue](https://github.com/alphagov/tech-docs-gem/issues/254), possible fixes can be found [here](#issue-with-ffi-on-osx-mohave).
```
richardTowers, line 37:
```suggestion
If you experience [this issue with ffi](https://github.com/alphagov/tech-docs-gem/issues/254), possible fixes can be found [here](#issue-with-ffi-on-osx-mohave).
```

README.md
```diff
@@ -63,6 +67,16 @@ To run the tests and see the results in your terminal, run:
 bundle exec rake jasmine:ci
 ```
 
+## Issue with FFI on OSX Mohave
```
richardTowers, line 70: Nitpick, but I think apple anglicise Mohave as "Mojave" in the name of their OS.

README.md
jonathanglassman: Can we make these links more descriptive and accessible? For example:

If you experience [the FFI gem issue for Mojave users](https://github.com/alphagov/tech-docs-gem/issues/254), you should refer to this [list of possible fixes](#issue-with-ffi-on-osx-mohave).

README.md
```diff
@@ -46,6 +48,8 @@ bundle exec middleman server
 
 See your website on `http://localhost:4567` in your browser.
 
+If you experience [this issue with ffi](https://github.com/alphagov/tech-docs-gem/issues/254), possible fixes can be found [here](#issue-with-ffi-on-osx-mohave).
```
jonathanglassman, line 51: If you experience the [FFI gem issue for Mojave users](https://github.com/alphagov/tech-docs-gem/issues/254), you should refer to this [list of possible fixes](#issue-with-ffi-on-osx-mohave).

README.md
```diff
@@ -34,6 +34,8 @@ gem 'govuk_tech_docs', path: '../tech-docs-gem'
 
 To preview your documentation changes locally, see the [Tech Docs Template documentation on previewing your documentation](https://tdt-documentation.london.cloudapps.digital/create_project/preview/#preview-your-documentation).
 
+If you experience [this issue with ffi](https://github.com/alphagov/tech-docs-gem/issues/254), possible fixes can be found [here](#issue-with-ffi-on-osx-mohave).
```
jonathanglassman, line 37: Could we make this more descriptive, accessible, and active? For example:

"If you experience the [FFI gem issue for Mojave users](https://github.com/alphagov/tech-docs-gem/issues/254), you should refer to this [list of possible fixes](#issue-with-ffi-on-osx-mohave)."

README.md
```diff
@@ -63,6 +67,16 @@ To run the tests and see the results in your terminal, run:
...
+## Issue with FFI on OSX Mojave
+
+Users on OSX Mojave (10.14) may get this error when running `bundle exec middleman serve` on apps that use this gem.
+
+There are 3 different ways to solve this, depending on what is possible for you. Listed here as best to worst:
+
+1. upgrade to macOS 10.15 (Catalina) or higher
+2. tell rubygems not to use the system ffi (gem inst ffi -- --disable-system-libffi)
+3. pin the ffi version back to 1.12.2 in the Gemfile of your app
+
```
jonathanglassman, line 79: Recommend we make the numbered bullets in line with style, and be more explicit in our instruction. For example:

"There are 3 possible ways to solve this. The following list is ranked from best to worst.

1. Upgrade to macOS 10.15 (Catalina) or higher. 
2. Tell rubygems not to use the system ffi by running `gem inst ffi -- --disable-system-libffi`
3. Pin the ffi version back to 1.12.2 by editing the Gemfile of your app.

I'm not sure how correct this is, for example, when and where do you run the `gem inst` command? But I would recommend being more explicit.

jonathanglassman (changes requested):
As per my comments. Happy to discuss.

README.md
tombye: I think these work as you wrote them. One tiny point: don't list items need to be lowercase? I'm going by [the GOVUK styleguide on bullets points and steps](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#bullet-points-steps).

tombye: Sorry, to answer your question, you'd run that command after running `bundle exec middleman serve` and it throwing the error mentioned. It would re-install the gem but with the `--disable-system-libffi` flag set to configure the installation process.

richardTowers: Looks like `gem inst` is a secret shorthand alias for gem install 😅 

```
% ▶ gem inst -h | head -n 1
Usage: gem install GEMNAME [GEMNAME ...] [options] -- --build-flags [options]
```

The long form (gem install) might be less confusing.

richardTowers: `gem i` and `gem ins` also work, but `gem in` is apparently ambiguous (between `gem info` and `gem install`)

jonathanglassman: Ah, I was thinking of numbered steps rather than bullets. I guess these aren't steps in a process, they are options. Maybe we could make them unordered bullets e.g.

"There are 3 possible ways to solve this. From best to worst, you can:

- upgrade to macOS 10.15 (Catalina) or higher
- tell rubygems not to use the system ffi by running `gem inst ffi -- --disable-system-libffi` in the command line when the error shows
- pin the ffi version back to 1.12.2 by editing the Gemfile of your app"

tombye: @jonathanglassman I've plugged in those changes and changed `gem inst` to `gem install` (thanks for the explanation @richardTowers) so marking this as resolved but feel free to undo that if anything outstanding/wrong.

jonathanglassman (approved):

PR #258, opened 2021-09-10
Bump Node version to latest LTS release

lfdebrux:
We want to make sure developers for this repo are using an up-to-date version of Node.

This commit updates the version of Node in the `.nvmrc` file to the current active long term support version, 14.

Note this also affects the version of Node used to build the Ruby Gem, but it shouldn't affect the contents of the gem in practice, because Node is only used to download the `govuk-frontend` package from npm.

Therefore this change doesn't need to go into the changelog or result in a new release.


owenatgov (approved):
1️⃣ 4️⃣

PR #259, opened 2021-09-10
Remove typeahead search functionality

zapthedingbat:

This is a fix for an accessibility issue related to displaying search results as the user types. There's a longer discussion in PR #239 and the referenced google sheet. The collusion in the immediate term seem to be:
> Quick fix could be as simple as turning off search-as-you-type - tech side would need to check with accessibility team this is WCAG-compliant.

---

> The issue
The remaining accessibility issue is essentially due to the TDT search results popping up over page content. DAC’s description of the issue is in cell B22 of the [TechDocs Template development spreadsheet](https://www.google.com/url?q=https://docs.google.com/spreadsheets/d/1s53eGMBj2hdTVLRitwi673R-_uERWh0kOxOp9QsOVg0/edit%23gid%3D0&sa=D&source=calendar&ust=1631715795418131&usg=AOvVaw3Bnc4gIfFvIz-cEc2P2XZF), and the discussion so far is in tech-doc-gems PR #239.
We’ve said we’ll fix this issue by 30 September in the TDT accessibility statement.
> ***
> Quick fix could be as simple as turning off search-as-you-type - tech side would need to check with accessibility team this is WCAG-compliant.

---

> "Issue ID: DAC_Search_Results_Dialog_01
> 
> Page: 37
> 
> 
> 
> 
> WCAG 2.1 (A) 
> - 3.2.2 On Input - https://www.w3.org/WAI/WCAG21/Understanding/on-input.html
>- 2.4.3 Focus Order - https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html"


m-green:
Thanks for raising this - this has been superseded by the PR that @lfdebrux has linked to, so closing this now.

PR #263, opened 2021-09-21
Change search to only show results after submit

### LINKED ISSUE #262: Search user experience is not accessible
From a recent DAC audit:

>
>
>
> WCAG 2.1 (A)
>   - 3.2.2 On Input - https://www.w3.org/WAI/WCAG21/Understanding/on-input.html
>   - 2.4.3 Focus Order - https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html

---

There are a couple of possible solutions to this:

1. We could go with the solution outlined by DAC and make the search results act more like a modal dialog, however as this [comment by @36degrees suggests this might not be a good model](https://github.com/alphagov/tech-docs-gem/pull/239#issuecomment-872992709).

2. @colinbm has raised PR #239 which removes the live updating, leaves the content visible, and reduces the content's opacity when the search results are visible, which may be an improvement from an accessibility standpoint but needs refinement and testing with a screenreader.

3. Make the search results behave less like a dialog and instead they just push down the rest of the content, as suggested by @selfthinker.

4. We could remove the live search functionality altogether - so no search suggestions, and no search dialog, just a normal form submission and a separate search results page, as suggested by @m-green. @zapthedingbat has raised PR #259 which does part of this, but creating a separate search results page would probably be a weeks worth extra work.

Option 4 is probably the most accessible and most guaranteed to work. It was also suggested by @selfthinker at the GDS accessibility clinic.

---

> NOTE: We want to resolve this issue by September 30th.

lfdebrux:
Closes #262, hopefully.

The search user experience is now more accessible for screenreader users.

Users can no longer see search results as you type, they must hit enter or click the search button to see their results.

Adds a new page, `/search`, which is used to show search results. The JavaScript previously used to search the index and render results is still used, it has just been limited in scope to only working on this page. The search input is now progressively enhanced to point to the search page.

The intention is to try and create a more 'traditional' browser experience, which will make the user experience with assistive tech better.

This is not a breaking change for consumers of the gem, but it is a major change of the search user experience. Previously the search results would appear immediately as the user types, however now the user must hit enter or click the search button to see the search results. 

## Preview

I've deployed a version of the GOV.UK Frontend tech docs with this branch of govuk_tech_docs to make it easier to preview the changes in this pull request.

You can access the preview at https://deploy-preview-147--govuk-frontend-docs-preview.netlify.app/search/.

If you are a consumer of the `govuk_tech_docs` gem you should also try this version locally by changing the `Gemfile` to use this branch:

```diff
--- a/Gemfile
+++ b/Gemfile
-gem "govuk_tech_docs"
+gem "govuk_tech_docs", git: "https://github.com/alphagov/tech-docs-gem.git", branch: "ldeb-search-page"
```

### Appearance of search results

| | Before | After |
|---|---|---|
| Chrome, desktop | search-results--before--chrome-desktop | search-results--after--chrome-desktop |
| Safari, iPhone 8 | search-results--before--safari-iphone8 | search-results--after--safari-iphone8 |

### Appearance of new search page

If the user visits the search page without a search query, a brief help page appears. If the user does not have JavaScript enabled the search page will not work, so there is also a help message that is visible only when JavaScript is not running.

| | |
|---|---|
| Chrome, desktop | search-page--chrome-desktop |
| Safari, iPhone 8 | search-page--safari-iphone8 |
| Safari, desktop, no JavaScript | search-page--safari-desktop-no-js |

---

Work remaining:

- [x] Fix/add tests
- [x] Try and remove/push down the footer on the search page
- [x] Show search input in mobile viewport
- [x] Test with different assistive tech and browser combinations
- [x] Take to the accessibility clinic
- [x] Add to commit messages/release notes
- [x] Content review
- [ ] (stretch goal) Clean up and refactor the Sass and JS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+### Breaking changes
+
+The search user experience is now more accessible for screenreader users. Search results are now on a separate page instead of a modal window.
+
+Users can no longer see search results as they type. They must press the Return key or select the search button to see the search results page.
+
+This was added in [pull request #263: Change search to only show results after submit](https://github.com/alphagov/tech-docs-gem/pull/263).
+
+### Fixes
+
 - [#265: Fix mark styles in Windows High Contrast Mode](https://github.com/alphagov/tech-docs-gem/pull/265)
 
 ## 2.4.3
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -190,7 +190,7 @@
     function updateTitle (text) {
       if (typeof text === 'undefined') {
         var count = results.length
-        var resultsText = count + ' results'
+        var resultsText = 'Search - ' + query + ' - ' + count + ' results'
       }
       $searchResultsTitle.text(text || resultsText)
     }
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -31,7 +31,7 @@
       changeSearchLabel()
 
       // Only do searches on the search page
-      if (window.location.pathname === '/search') {
+      if (window.location.pathname.match(/\/search(\/|\/index.html)?$/)) {
         $html.addClass('has-search-results-open')
 
         if (window.location.search) {
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -13,11 +13,9 @@
     var $searchResults
     var $searchResultsTitle
     var $searchResultsWrapper
-    var $searchResultsClose
     var $searchHelp
     var results
     var query
-    var queryTimer
     var maxSearchEntries = 20
 
     this.start = function start ($element) {
@@ -27,16 +25,25 @@
       $searchResultsWrapper = $element.find('.search-results')
       $searchResults = $searchResultsWrapper.find('.search-results__content')
       $searchResultsTitle = $searchResultsWrapper.find('.search-results__title')
-      $searchResultsClose = $searchResultsWrapper.find('.search-results__close')
       $searchHelp = $('#search-help')
       $tocNav = $('#toc')
       s.downloadSearchIndex()
-      attach()
+      changeSearchAction()
       changeSearchLabel()
+
+      // Only do searches on the search page
+      if (window.location.pathname === '/search' && window.location.search) {
+        query = getQueryFromURL()
+        if (query) {
+          $searchInput.val(query)
+          doSearch(query)
+          doAnalytics()
+        }
+      }
     }
 
     this.downloadSearchIndex = function downloadSearchIndex () {
-      updateTitle('Loading search index')
+      updateTitle('Loading search results')
       $.ajax({
         url: '/search.json',
         cache: true,
@@ -50,46 +57,41 @@
       })
     }
 
-    function attach () {
-      // Search functionality on search text input
-      $searchInput.on('input', function (e) {
-        e.preventDefault()
-        query = $(this).val()
-        s.search(query, function (r) {
-          results = r
-          renderResults(query)
-          updateTitle()
-        })
-        if (window.ga) {
-          window.clearTimeout(queryTimer)
-          queryTimer = window.setTimeout(sendQueryToAnalytics, 1000)
-        }
-      })
+    function changeSearchAction () {
+      // We need JavaScript to do search, so if JS is not available the search
+      // input sends the query string to Google. This JS function changes the
+      // input to instead send it to the search page.
+      $searchForm.prop('action', '/search')
+      $searchForm.find('input[name="as_sitesearch"]').remove()
+    }
 
-      // Set focus on the first search result instead of submiting the search
-      // form to Google
-      $searchForm.on('submit', function (e) {
-        e.preventDefault()
-        showResults()
-        var $firstResult = $searchResults.find('.search-result__title a').first()
-        if ($firstResult.length) {
-          $firstResult.focus()
-        } else {
-          // if there are no results, show the "0 results" state
-          results = []
-          updateTitle()
-        }
-      })
+    function changeSearchLabel () {
+      $searchLabel.text('Search this documentation')
+    }
 
-      // Closing the search results, move focus back to the search input
-      $searchResultsClose.on('click', function (e) {
-        e.preventDefault()
-        $searchInput.focus()
-        hideResults()
-      })
+    function getQueryFromURL () {
+      var query = decodeURIComponent(
+        window.location.search
+          .match(/q=([^&]*)/)[1]
+          .replace(/\+/g, ' ')
+      )
+      return query
+    }
 
-      // Attach analytics events to search result clicks
+    function doSearch (query) {
+      s.search(query, function (r) {
+        results = r
+        renderResults(query)
+        updateTitle()
+      })
+    }
+
+    // TODO: remove this and sendQueryToAnalytics in a future breaking release
+    function doAnalytics () {
       if (window.ga) {
+        sendQueryToAnalytics()
+
+        // Attach analytics events to search result clicks
         $searchResults.on('click', '.search-result__title a', function () {
           var href = $(this).attr('href')
           ga('send', {
@@ -101,15 +103,6 @@
           })
         })
       }
-
-      // When selecting navigation link, close the search results.
-      $('#toc').on('click', 'a', function (e) {
-        hideResults()
-      })
-    }
-
-    function changeSearchLabel () {
-      $searchLabel.text('Search this documentation')
     }
 
     function getResults (query) {
@@ -124,7 +117,6 @@
 
     this.search = function search (query, callback) {
       if (query === '') {
-        hideResults()
         return
       }
       showResults()
@@ -204,13 +196,6 @@
         .attr('aria-hidden', 'false')
       $html.addClass('has-search-results-open')
       $tocNav.addClass('search-results-open')
-    }
-
-    function hideResults () {
-      $searchResultsWrapper.removeClass('is-open')
-        .attr('aria-hidden', 'true')
-      $html.removeClass('has-search-results-open')
-      $tocNav.removeClass('search-results-open')
     }
 
     function sendQueryToAnalytics () {
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -6,7 +6,6 @@
   Modules.Search = function Search () {
     var s = this
     var $html = $('html')
-    var $tocNav
     var $searchForm
     var $searchLabel
     var $searchInput
@@ -26,18 +25,22 @@
       $searchResults = $searchResultsWrapper.find('.search-results__content')
       $searchResultsTitle = $searchResultsWrapper.find('.search-results__title')
       $searchHelp = $('#search-help')
-      $tocNav = $('#toc')
+
       s.downloadSearchIndex()
       changeSearchAction()
       changeSearchLabel()
 
       // Only do searches on the search page
-      if (window.location.pathname === '/search' && window.location.search) {
-        query = getQueryFromURL()
-        if (query) {
-          $searchInput.val(query)
-          doSearch(query)
-          doAnalytics()
+      if (window.location.pathname === '/search') {
+        $html.addClass('has-search-results-open')
+
+        if (window.location.search) {
+          query = getQueryFromURL()
+          if (query) {
+            $searchInput.val(query)
+            doSearch(query)
+            doAnalytics()
+          }
         }
       }
     }
@@ -194,8 +197,6 @@
     function showResults () {
       $searchResultsWrapper.addClass('is-open')
         .attr('aria-hidden', 'false')
-      $html.addClass('has-search-results-open')
-      $tocNav.addClass('search-results-open')
       $searchHelp.attr('hidden', 'true')
     }
 
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -14,6 +14,7 @@
     var $searchResultsTitle
     var $searchResultsWrapper
     var $searchResultsClose
+    var $searchHelp
     var results
     var query
     var queryTimer
@@ -27,6 +28,7 @@
       $searchResults = $searchResultsWrapper.find('.search-results__content')
       $searchResultsTitle = $searchResultsWrapper.find('.search-results__title')
       $searchResultsClose = $searchResultsWrapper.find('.search-results__close')
+      $searchHelp = $('#search-help')
       $tocNav = $('#toc')
       s.downloadSearchIndex()
       attach()
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -196,8 +196,7 @@
     }
 
     function showResults () {
-      $searchResultsWrapper.addClass('is-open')
-        .attr('aria-hidden', 'false')
+      $searchResultsWrapper.removeAttr('hidden')
       $searchHelp.attr('hidden', 'true')
     }
 
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -40,6 +40,7 @@
             $searchInput.val(query)
             doSearch(query)
             doAnalytics()
+            document.title = query + ' - ' + document.title
           }
         }
       }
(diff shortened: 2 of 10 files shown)


lib/assets/javascripts/_modules/search.js
```diff
@@ -35,6 +35,7 @@
       if (window.location.pathname === '/search' && window.location.search) {
         query = getQueryFromURL()
         doSearch(query)
+        doAnalytics()
```
ChrisBAshton: This 'restore' commit should be fixed up to the previous commit, as there's little point in removing code only to add it back in straight away.

lib/assets/javascripts/_modules/search.js
```diff
@@ -196,6 +196,7 @@
         .attr('aria-hidden', 'false')
       $html.addClass('has-search-results-open')
       $tocNav.addClass('search-results-open')
+      $searchHelp.attr('hidden', 'true')
```
ChrisBAshton, line 200: What happens if the user clears the search input field? Presumably the help text stays `aria-hidden`, when it should actually be available to assistive tech again?

lib/source/search/index.html.erb
```diff
@@ -6,7 +6,8 @@ hide_in_navigation: true
 ---
 
 <div id="search-help" class="search-help">
-  <h1>Search</h1>
+  <%# strange id is to avoid clash with search input id #%>
+  <h1 id="how-to-search">Search</h1>
```
ChrisBAshton, line 10: This looks good - thanks for the comment 👍

ChrisBAshton (commented):
Thanks for taking this on, @lfdebrux 👏 I haven't tried it locally yet (that's next on my list), but the screenshots and description of the new behaviour are helpful. I've just got a couple of notes below:

lib/assets/javascripts/_modules/search.js
lfdebrux: If they clear the search input nothing will happen, as we've removed the typeahead, so I think this is fine?

zapthedingbat (approved):
This looks good. @selfthinker suggested that we either, 1) try to make the modal more constrained to stop the user interacting with the rest of the page, or 2) push the results down into the page like the rest of the content. My feeling is that this option (2) is less prone to confusion and errors, because we're not trying to manipulate the native focus behaviours of the browser or screen readers.

m-green:
Thank you @selfthinker for reviewing!

4 minor accessibility fixes needed.

1. Title is not unique - suggest it starts with `{search term} - Search` instead of just `Search - ` (same as GOV.UK).
2. Similar the first heading is not unique - suggest we use 'Search - {search term} - X results'.
3. Jaws missed the first heading - need to make this a proper h1.
4. Headings are overriding the default aria role. @selfthinker suggested moving the assertive/alert aria roles in a `<span>` inside the `<h2>`

Let me know if I got anything wrong!

selfthinker:
Just to clarify, Jaws missed the first heading because of the aria role, not because it wasn't an h1.

m-green:
Thanks for clarifying @selfthinker!

lfdebrux:
Thanks for the review and write-up @selfthinker , @m-green 

I've pushed extra commits addressing the issues identified.

lib/source/search/index.html.erb
```diff
@@ -0,0 +1,26 @@
...
+prevent_indexing: true
+hide_in_navigation: true
+---
+
+<div id="search-help" class="search-help">
+  <%# strange id is to avoid clash with search input id #%>
+  <h1 id="how-to-search">Search</h1>
+
+  <p class="govuk-body search-help__no-js">
+    This page requires JavaScript to perform searches. Please enable JavaScript, or use the search input to search using Google.
```
m-green, line 13:
```suggestion
You need to turn on JavaScript in your browser to search this site. You can either:

<ul>
    <li>turn on JavaScript in your browser</li>
    <li>reenter your search term and select **Search** again, to search this site using Google</li>
</ul>
```

lib/source/search/index.html.erb
```diff
@@ -0,0 +1,26 @@
...
+  </p>
+
+  <p class="govuk-body">
```
m-green, line 16: If I understand right, I think we can remove this paragraph? Seems to duplicate the one above.

m-green (commented):
Looks good! Just some suggested content tweaks.

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+### Breaking changes
+
+The search user experience is now more accessible for screenreader users.
```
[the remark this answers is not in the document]
lfdebrux: My understanding from the [DAC finding](https://github.com/alphagov/tech-docs-gem/issues/262) was that the issue with the search modal mainly affected screen reader users, I could be wrong though.

lib/source/search/index.html.erb
lfdebrux: Do you want 'Search' to be bold, or surrounded by asterisks?

CHANGELOG.md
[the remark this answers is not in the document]
m-green: Yes I think you're right. All good then - just my suggested addition above to explain the change a bit more.

lib/source/search/index.html.erb
lfdebrux: This paragraph currently appears whether or not the user has JavaScript enabled; I guess we could make it hidden until some JS unhides it?

lib/source/search/index.html.erb
m-green: Ah oops, I was in Markdown mode. Yep that should be bold, sorry.

lib/source/search/index.html.erb
m-green: Do you know what triggers it to appear if you have JS on? I could only get it to show up if I switched JS off.
(Btw if this paragraph was already part of the search before this PR, please leave as is - don't want to inadvertently increase the scope of the PR)

lfdebrux: Ah, sorry, I was confused, I thought your original comment was referring to the paragraph reading 'Search the documentation using the search input.'.

lfdebrux: I'm not sure what you mean by this paragraph duplicating the one above.

m-green: Ah yes it was! I just can't get that second paragraph to show up, unless I turn JS off. So wondered in what circumstances it shows up.

lfdebrux: Can you see it if you visit https://deploy-preview-147--govuk-frontend-docs-preview.netlify.app/search/?

It appears when the user clicks the search button but has not entered a search query.

lib/source/search/index.html.erb
lfdebrux: Hmm, I'm not sure about 'reenter... and select Search again'; in most cases readers should only see this content if they've not entered a search term.

(They could also reach this content if they visit the search page with a search query already in the URL, but that seems unlikely to me).

m-green: Ah - I thought they would see the "This page requires JavaScript..." paragraph only if they searched with JS off? When I search without a search term, I get the other paragraph ("Search the documentation using the search input.")

lib/source/search/index.html.erb
m-green: Ah yes I can see it! In that case we should probably tweak this a bit to be clearer. So I'd suggest replacing "Search the documentation using the search input." with "Enter a search term in the search box, then select **Search**".

lfdebrux: I'm not sure about 'select **Search**', as the search button doesn't actually say search on it... what about 'click the search button', or something similar?

lib/source/search/index.html.erb
lfdebrux: Sorry, maybe I should have made it clearer in the search commit or somewhere else... my thinking for this content (both content about JavaScript and without) is that it would be there in case the user somehow found the search page without entering a search query; the most common way I think this would happen is by clicking the search button without entering anything into the search box. If they had entered a search term they wouldn't need the help text, was my thinking.

I'm happy to have a call to discuss further, if that would be helpful.

lib/source/search/index.html.erb
m-green: Yep that works! Just use 'select' instead of 'click' as it's more accessible: https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#click

lib/source/search/index.html.erb
m-green: Thanks for taking me through this. So if I've understood right, I think this would work for the 'first' paragraph?

```
You need to turn on JavaScript in your browser to search this site. You can either:

<ul>
    <li>turn on JavaScript in your browser</li>
    <li>search this site using Google</li>
</ul>
```

Then (if they searched for nothing AND JS is off) they'll also see the 'second' paragraph, which would be:

```
Enter a search term in the search box, then select the search button.
```

Let me know if I got anything wrong!

lib/source/search/index.html.erb
lfdebrux: Ah cool, didn't know that!

lib/source/search/index.html.erb
lfdebrux: Yep, that looks good!

I've made those changes now, please take another look to check that I've done it right! 😅

m-green: Perfect, thanks @lfdebrux!

lib/assets/javascripts/_modules/search.js
ChrisBAshton: This is still unresolved: could it be fixed up to the previous commit? Usually we try to avoid removing code and then re-adding it in, in the same PR. Better to consolidate it into one good commit. I appreciate the `doAnalytics` function is new, so it's not _exactly_ the same, but that can be added in to the previous commit too.

lfdebrux: Oh, yep, fixed that now.

ChrisBAshton (commented):
I've now had a chance to try this out locally. Nice job - it works really well!

The new nav on the left hand side was a bit of a cause of confusion though. I get that the `Search` link takes you to the search help text, and the `Results` link takes you to the results (if there are any), but there's no visual feedback that this is what it's doing, as I guess it was primarily for assistive tech users.

> Screenshot 2021-10-05 at 10 58 42

I think having both links visible at the same time is a bit of a strange one. When there are no results, the `Results` link takes you nowhere, and when there are results, the `Search` link takes you nowhere. Or rather, it takes you to `#how-to-search`, but this is within a `<div>` that has `aria-hidden` on it, so presumably the anchor won't resolve for assistive tech users in these cases.

I think a better solution would be to either remove the links altogether, or, if it is an accessibility requirement, perhaps combine them into one general "Skip to main content" link. This would link to the ID of the container that would, depending on context, either have help text, or results. I would then make this general skip link visible only on focus, i.e. invisible by default to sighted users. Do you have any thoughts on this?

Would be good to get a frontend dev's eyes on this too 👀 . Otherwise, code-wise, I think this is good to go 👍

lfdebrux:
> I've now had a chance to try this out locally. Nice job - it works really well!

Fantastic, thanks very much!

> The new nav on the left hand side was a bit of a cause of confusion though ... I think a better solution would be to either remove the links altogether

Ah, I'm guessing you're looking at [govuk-developer-docs, which does not using the multipage nav](https://github.com/alphagov/govuk-developer-docs/blob/main/config/tech-docs.yml)?

Removing the single page table of contents sidebar navigation for the sidebar seems fine to me, I will add a commit to do this.

lib/assets/stylesheets/modules/_search.scss
```diff
@@ -52,48 +52,44 @@ $input-size: 40px;
...
-      @include govuk-link-common;
-      @include govuk-link-style-no-visited-state;
-    }
+  ul {
+    list-style: none;
+    padding: 0;
+    margin: 0;
   }
+}
 .search-results__inner {
```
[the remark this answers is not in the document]
lfdebrux: Hmm, I see your point, but I'm reluctant to make a change without input for a designer... I'd want to make sure any change would be making things better rather than worse! I think it's important to keep a certain amount of right padding, but that amount might be different at different viewports.

I think the best thing for now would be to create a new issue about this, is that something you would be happy to do? Sorry to be difficult!

lfdebrux:
> > The new nav on the left hand side was a bit of a cause of confusion though ... I think a better solution would be to either remove the links altogether
> 
> Ah, I'm guessing you're looking at [govuk-developer-docs, which does not using the multipage nav](https://github.com/alphagov/govuk-developer-docs/blob/main/config/tech-docs.yml)?
> 
> Removing the single page table of contents sidebar navigation for the sidebar seems fine to me, I will add a commit to do this.

Ah, this is going to be harder than I thought... the naive way to have an empty sidebar navigation is to override [the sidebar content in the core erb template](https://github.com/alphagov/tech-docs-gem/blob/master/lib/source/layouts/core.erb), however for mobile devices this either leaves a table of contents button that does nothing, or omits the search box entirely.

So making this change the proper way would involve massively rewriting the core template, which will take time and be risky. I guess the hacky way to do it would be to do it with JavaScript; this will be slightly easier, but not so great maintenance wise.

lib/assets/stylesheets/modules/_search.scss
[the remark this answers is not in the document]
ChrisBAshton: Totally understood. I'll raise an issue after this PR is merged (as it isn't an issue until then!).

ChrisBAshton (approved):
Approving, with the caveat that I think a frontend dev should also approve before this is merged.

Nice work ⭐

lib/assets/javascripts/_modules/search.js
```diff
@@ -6,35 +6,47 @@
...
       $searchLabel = $element.find('.search__label')
-      $searchResultsWrapper = $element.find('.search-results')
+      $searchResultsWrapper = $('#search-results')
       $searchResults = $searchResultsWrapper.find('.search-results__content')
       $searchResultsTitle = $searchResultsWrapper.find('.search-results__title')
-      $searchResultsClose = $searchResultsWrapper.find('.search-results__close')
-      $tocNav = $('#toc')
+      $searchHelp = $('#search-help')
+
       s.downloadSearchIndex()
```
36degrees, line 29: Would it make sense to now only download the search index if we're on the search page? If the search JSON has poor cache headers (which sadly is probably quite likely) then it'll be re-requested when the user hits the search page anyway.

36degrees (commented):
I think broadly speaking these changes look OK. I'm still a little unsure about the general idea of trying to mimic a server-side search, but I know this has tested well at the accessibility clinic so happy to defer.

I do think the code is getting a little unwieldy and hard to follow, but we should look at refactoring this later (ideally having improved the test coverage).

There are a couple of minor comments that could do with looking at before I approve.

lib/assets/javascripts/_modules/search.js
lfdebrux: Mmm, you're right, we might as well avoid the extra requests. If in future we make it easier to have better caching we can always change it back.

lib/assets/javascripts/_modules/search.js
```diff
@@ -48,46 +60,44 @@
...
+      var query = decodeURIComponent(
+        window.location.search
+          .match(/q=([^&]*)/)[1]
+          .replace(/\+/g, ' ')
+      )
+      if (query) {
+        $searchInput.val(query)
+      }
+      return query
+    }
```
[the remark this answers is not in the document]
lfdebrux: Good point. Will fix.

lib/assets/javascripts/_modules/search.js
36degrees: Sounds good. At some point it might also make sense to try and dynamically preload the search index when the user interacts with the search box, but again that only really makes sense if we know that any caching control will allow the cached index to be re-used on the next page, otherwise we're back to loading it twice.

lfdebrux: I've added a separate commit to make this change, it also required a tweak to the tests.

lfdebrux:
Thank you to everyone who reviewed this, especially @zapthedingbat, @ChrisBAshton, @m-green, @selfthinker, and @36degrees.

I'm aware that there are one or two unresolved comments, we've agreed on Slack though that we can release these changes as-is and fix the outstanding issues in later releases; we're feeling good that these changes are an improvement accessibility-wise and will help users, as well as us and service teams.

lib/assets/stylesheets/modules/_search.scss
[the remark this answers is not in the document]
ChrisBAshton: I've now raised https://github.com/alphagov/tech-docs-gem/issues/269.

PR #273, opened 2021-10-20
Don't construct `report_issue_url` through string manipulation

jamietanna:
Instead of constructing the URL for reporting an issue by manipulating
the raw string, we can instead use the `URI.encode_www_form`[0] method.

This ensures that we're constructing the URL as a well-formed, encoded
set of parameters, and has the added benefit of it being much clearer to
read, too.

[0]: https://stackoverflow.com/a/11251654

### DIFF
--- a/lib/govuk_tech_docs/contribution_banner.rb
+++ b/lib/govuk_tech_docs/contribution_banner.rb
@@ -20,12 +20,18 @@ module GovukTechDocs
 
     def report_issue_url
       url = config[:source_urls]&.[](:report_issue_url)
+      params = {
+        body: "Problem with '#{current_page.data.title}' (#{config[:tech_docs][:host]}#{current_page.url})",
+      }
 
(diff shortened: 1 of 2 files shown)


m-green:
Hi @jamietanna! Thanks for submitting this. We just spotted that this changes the text of the link from “Problem with …” to “Feedback to …“ - was that intentional?

lfdebrux (approved):
These changes look good to me, and much easier to read! Thanks!

jamietanna:
Sorry @m-green no that wasn't expected - that was copy-paste error from where we're using it, because we've tweaked the wording

m-green:
> Sorry @m-green no that wasn't expected - that was copy-paste error from where we're using it, because we've tweaked the wording

No worries @jamietanna - thought that might have been the case! I'll approve now.

m-green (approved):

PR #280, opened 2021-11-03
Fix CI

lfdebrux:
A v4.4.1 version of middleman was released yesterday that seems to be broken (see for instance failed tests on main branch https://github.com/alphagov/tech-docs-gem/runs/4095543211?check_suite_focus=true).

This PR adds a requirement to the gemspec that rejects this version, so our users won't install it. It also fixes our tests.

### DIFF
--- a/govuk_tech_docs.gemspec
+++ b/govuk_tech_docs.gemspec
@@ -34,7 +34,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "autoprefixer-rails", "~> 10.2"
   spec.add_dependency "chronic", "~> 0.10.2"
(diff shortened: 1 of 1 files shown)


lfdebrux:
This PR is no longer needed, as the release v4.4.1 was yanked from RubyGems by the author and replaced with v4.4.2

https://github.com/middleman/middleman/commit/5647a1b87a060f551d5936f3482ec73e2b73cad2

I'm not sure what was wrong with v4.4.1, but our tests now pass on the main branch, so I'm happy

https://github.com/alphagov/tech-docs-gem/runs/4115161992?check_suite_focus=true

PR #197, opened 2020-09-30
Assets: Support Github pages / http_prefix

odlp:
Hello 👋 

This PR fixes asset URLs for sites built for Github pages using the default URL (like the [API catalogue](https://github.com/alphagov/api-catalogue)), in which paths are prefixed with the repo name.

## Changes

- Allow `$govuk-assets-path` to be overriden by implementors
- Use Sprockets URL helpers to ensure image backround URLs in CSS account for Middleman's `http_prefix` setting

## Context

For repo-specific Github pages, the site is deployed with the repo name prefixed in the path, e.g. `https://alphagov.github.io/api-catalogue/`. We'd like to use Middleman's [`http_prefix`](https://rubydoc.info/github/middleman/middleman/Middleman/Application#http_prefix-instance_method) option to account for this path prefix, which will allow us to simplify and automate  our build process.

Our Middleman config looks like this:

```ruby
# Without prefix for 'middleman serve'
set(:govuk_assets_path, "/assets/govuk/assets/")

# Add '/api-catalogue/' for 'middleman build', for Github Pages compatibility
configure :build do
  set(:build_dir, "build/api-catalogue")
  set(:http_prefix, "/api-catalogue/")
  set(:govuk_assets_path, "/api-catalogue/assets/govuk/assets/")
end
```

With this PR all asset URLs correctly include the `/api-catalogue/` prefix, and we can avoid using Middleman's `activate :relative_assets` feature.

## References

- [middleman-sprockets `asset_path` / `image-url` helpers](https://github.com/middleman/middleman-sprockets/blob/5e5197c3e4049845c446dc2b30d0169f2b3c6070/features/middleman_helpers.feature#L3)
- GOV.UK Frontend [using `!default` to allow overrides](https://github.com/alphagov/govuk-frontend/blob/063cd8e2470b62b824c6e50ca66342ac7a95d2d8/src/govuk/settings/_assets.scss#L14)
- Semi-related PR which [fixes search with `http_prefix` configued](https://github.com/alphagov/tech-docs-gem/pull/196)

PR #196, opened 2020-09-29
Search: Support Github pages / http_prefix

odlp:
Hello 👋 . I'm working on the [API catalogue](https://github.com/alphagov/api-catalogue) with the DSA team. We use the marvellous `tech-docs-gem` to build the site, which is then currently deployed to Github pages.

## Context

For repo-specific Github pages, the site is deployed with the repo name prefixed in the path, e.g. `https://alphagov.github.io/api-catalogue/`. We'd like to use Middleman's [`http_prefix`](https://rubydoc.info/github/middleman/middleman/Middleman/Application#http_prefix-instance_method) option to account for this path prefix, which will allow us to simplify and automate  our build process.

Unfortunately with `http_prefix` configured the search doesn't work at the moment. Browsing `/api-catalogue/index.html` leads to a `404 Not found` for `/search.json` (the asset is actually located at `/api-catalogue/search.json`).

## Fix

Switches to use the `search_index_path` helper provided by the Middleman search gem itself, rather than hardcoding `/search.json`. This helper accounts for the `http_prefix` setting.

References:

- https://github.com/manastech/middleman-search#asset-pipeline
- https://github.com/manastech/middleman-search/blob/ed9e5254bb554988c755fc40238d9e9f67b20abe/lib/middleman-search/extension.rb#L29

PR #291, opened 2022-01-20
Support sites deployed on paths other than "/" (by generating relative links)

eddgrant:
## What are you trying to do?

Make it possible to generate a fully working site, in which all intra-site links (i.e. links within the site, between pages, to assets etc) are generated as relative links, so that the site can be deployed at a path other than "/".

## Identifying a user need

At HMRC we use the tech-docs-template to deploy multiple docs-as-code sites at different paths on a single domain e.g.

```
docs.tax.service.gov.uk/site-1
docs.tax.service.gov.uk/site-2
docs.tax.service.gov.uk/site-3
... etc
```

Without these changes the site links do not function correctly.

## Is it a change or a bug fix?

Change.

## What’s changed?

The changes are driven via standard Middleman configuration:

```yaml
set :relative_links, true
activate :relative_assets
```

When enabled, the above configuration causes the following changes to happen:

1. The Table of Contents is generated using relative links (824116a9f3be15c9641e9047d638dca84ff5ed85)
2. The Table of Contents expands, collapsed and is highlighted using relative links (a2dde29337afcb84c7bd19ac800b0cf074072af7)
3. The search function works using relative links (a2dde29337afcb84c7bd19ac800b0cf074072af7)
4. The `/api/pages.json` endpoint returns relative links, such that the `tech-docs-monitor` can correctly generate page URLs when raising Slack notifications (16396e70a8db310ccbaca8121fbe9eb2bdb484bc). This requires a companion `tech-docs-monitor` change for which the PR is [here](https://github.com/alphagov/tech-docs-monitor/pull/37).

### Testing

We are testing these changes at HMRC at the moment and haven't yet observed anything which has broken as a result of the changes. We don't believe the changes cause any regressions when building sites in the normal way using absolute links. However we'd be grateful for any additional testing people can do.

### DIFF
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -30,7 +30,7 @@
       searchIndexPath = $element.data('searchIndexPath')
       pathToSiteRoot = $element.data('pathToSiteRoot')
 
-      changeSearchAction(pathToSiteRoot)
+      changeSearchAction()
       changeSearchLabel()
 
       // Only do searches on the search page
@@ -65,7 +65,7 @@
       })
     }
 
-    function changeSearchAction (pathToSiteRoot) {
+    function changeSearchAction () {
       // We need JavaScript to do search, so if JS is not available the search
       // input sends the query string to Google. This JS function changes the
       // input to instead send it to the search page.
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -153,7 +153,7 @@
         var content = s.processContent(result.content, query)
         output += '<li class="search-result">'
         output += '<h3 class="search-result__title">'
-        const url = result.url.startsWith('/') ? result.url.slice(1) : result.url
+        var url = result.url.startsWith('/') ? result.url.slice(1) : result.url
         output += '<a href="' + pathToSiteRoot + url + '">'
         output += result.title
         output += '</a>'
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -153,8 +153,9 @@
         var content = s.processContent(result.content, query)
         output += '<li class="search-result">'
         output += '<h3 class="search-result__title">'
-        var url = result.url.startsWith('/') ? result.url.slice(1) : result.url
-        output += '<a href="' + pathToSiteRoot + url + '">'
+        var pagePathWithoutLeadingSlash = result.url.startsWith('/') ? result.url.slice(1) : result.url
+        var url = pathToSiteRoot.startsWith('.') ? pathToSiteRoot + pagePathWithoutLeadingSlash : '/' + pagePathWithoutLeadingSlash
+        output += '<a href="' + url + '">'
         output += result.title
         output += '</a>'
         output += '</h3>'
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -78,8 +78,7 @@
     }
 
     this.isOnSearchPage = function isOnSearchPage () {
-      // .* prefix is used to match sites which use relative links.
-      return Boolean(window.location.pathname.match(/.*\/search(\/|\/index.html)?$/))
+      return Boolean(window.location.pathname.match(/\/search(\/|\/index.html)?$/))
     }
 
     this.getQuery = function getQuery () {
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -16,6 +16,7 @@
     var results
     var query
     var maxSearchEntries = 20
+    var pathToSiteRoot
     var searchIndexPath
 
     this.start = function start ($element) {
@@ -27,8 +28,9 @@
       $searchResultsTitle = $searchResultsWrapper.find('.search-results__title')
       $searchHelp = $('#search-help')
       searchIndexPath = $element.data('searchIndexPath')
+      pathToSiteRoot = $element.data('pathToSiteRoot')
 
-      changeSearchAction()
+      changeSearchAction(pathToSiteRoot)
       changeSearchLabel()
 
       // Only do searches on the search page
@@ -40,7 +42,7 @@
           query = s.getQuery()
           if (query) {
             $searchInput.val(query)
-            doSearch(query)
+            doSearch(query, pathToSiteRoot)
             doAnalytics()
             document.title = query + ' - ' + document.title
           }
@@ -51,7 +53,7 @@
     this.downloadSearchIndex = function downloadSearchIndex () {
       updateTitle('Loading search results')
       $.ajax({
-        url: searchIndexPath,
+        url: pathToSiteRoot + (searchIndexPath.startsWith('/') ? searchIndexPath.slice(1) : searchIndexPath),
         cache: true,
         method: 'GET',
         success: function (data) {
@@ -63,11 +65,11 @@
       })
     }
 
-    function changeSearchAction () {
+    function changeSearchAction (pathToSiteRoot) {
       // We need JavaScript to do search, so if JS is not available the search
       // input sends the query string to Google. This JS function changes the
       // input to instead send it to the search page.
-      $searchForm.prop('action', '/search')
+      $searchForm.prop('action', pathToSiteRoot + 'search/index.html')
       $searchForm.find('input[name="as_sitesearch"]').remove()
     }
 
@@ -76,7 +78,8 @@
     }
 
     this.isOnSearchPage = function isOnSearchPage () {
-      return Boolean(window.location.pathname.match(/\/search(\/|\/index.html)?$/))
+      // .* prefix is used to match sites which use relative links.
+      return Boolean(window.location.pathname.match(/.*\/search(\/|\/index.html)?$/))
     }
 
     this.getQuery = function getQuery () {
@@ -88,10 +91,10 @@
       return query
     }
 
-    function doSearch (query) {
+    function doSearch (query, pathToSiteRoot) {
       s.search(query, function (r) {
         results = r
-        renderResults(query)
+        renderResults(query, pathToSiteRoot)
         updateTitle()
       })
     }
@@ -140,7 +143,7 @@
       callback(getResults(query))
     }
 
-    function renderResults (query) {
+    function renderResults (query, pathToSiteRoot) {
       var output = ''
       if (results.length === 0) {
         output += '<p>Nothing found</p>'
@@ -151,7 +154,8 @@
         var content = s.processContent(result.content, query)
         output += '<li class="search-result">'
         output += '<h3 class="search-result__title">'
-        output += '<a href="' + result.url + '">'
+        const url = result.url.startsWith('/') ? result.url.slice(1) : result.url
+        output += '<a href="' + pathToSiteRoot + url + '">'
         output += result.title
         output += '</a>'
         output += '</h3>'
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -17,7 +17,6 @@
     var query
     var maxSearchEntries = 20
     var pathToSiteRoot
-    var searchIndexPath
 
     this.start = function start ($element) {
       $searchForm = $element.find('form')
@@ -27,7 +26,6 @@
       $searchResults = $searchResultsWrapper.find('.search-results__content')
       $searchResultsTitle = $searchResultsWrapper.find('.search-results__title')
       $searchHelp = $('#search-help')
-      searchIndexPath = $element.data('searchIndexPath')
       pathToSiteRoot = $element.data('pathToSiteRoot')
 
       changeSearchAction()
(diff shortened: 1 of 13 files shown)


lfdebrux:
@eddgrant looks like there are some linter warnings, are you happy to fix these?

lib/assets/javascripts/_modules/search.js
```diff
@@ -74,7 +76,8 @@
     }
 
     this.isOnSearchPage = function isOnSearchPage () {
-      return Boolean(window.location.pathname.match(/\/search(\/|\/index.html)?$/))
+      // .* prefix is used to match sites which use relative links.
+      return Boolean(window.location.pathname.match(/.*\/search(\/|\/index.html)?$/))
```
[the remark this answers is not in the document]
eddgrant: That's a very good question! I thought it wasn't working without my amendment, however I wasn't aware that the match happened anywhere within the string. I will remove and test to see what happens.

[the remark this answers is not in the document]
eddgrant: Have tested this and you're absolutely right, the ".*" is unnecessary. Will remove.

lib/assets/javascripts/_modules/search.js
```diff
@@ -25,8 +26,9 @@
       $searchResults = $searchResultsWrapper.find('.search-results__content')
       $searchResultsTitle = $searchResultsWrapper.find('.search-results__title')
       $searchHelp = $('#search-help')
+      pathToSiteRoot = $element.data('pathToSiteRoot')
 
-      changeSearchAction()
+      changeSearchAction(pathToSiteRoot)
```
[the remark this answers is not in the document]
eddgrant: Good spot, have removed the variable from the function call.

spec/table_of_contents/helpers_spec.rb
```diff
@@ -175,7 +281,8 @@ def add_children(children)
       expect(subject.multi_page_table_of_contents(resources, current_page, config, current_page_html).strip).to eq(expected_multi_page_table_of_contents.strip)
     end
 
-    it "builds a table of contents from several page resources with a custom http prefix ocnfigured" do
+    #TODO: Assess whether this test is still relevant given that we are now generating relative links.
```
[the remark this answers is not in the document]
eddgrant: Great, have removed the TODO.

eddgrant:
I've just realised the `spec` task doesn't run the Jasmine tests (which are failing). I'm not familiar with Jasmine but I'll see if I can figure out what's going on there.

eddgrant:
Got it! I was using EC6 `const` variable declarations, which work fine in my browser, but I think Jasmine may be configured to build at a prior version to ES6 and didn't like the `const` declarations. Changing them to `var` declarations has done the trick, tests are now all passing.

lib/govuk_tech_docs/path_helpers.rb
```diff
@@ -0,0 +1,29 @@
...
+      resource_path
+    end
+
+    def path_to_site_root(config, page_path)
+      if config[:relative_links]
+        number_of_ascents_to_site_root = page_path.to_s.split("/").reject(&:empty?)[0..-2].length
+        ascents = number_of_ascents_to_site_root.zero? ? ["."] : number_of_ascents_to_site_root.times.collect { ".." }
+        path_to_site_root = ascents.join("/").concat("/")
+      else
+        path_to_site_root = "/"
```
lfdebrux, line 24: Can we include `config[:http_prefix]` here, for cases where the user is using that instead of relative links?

lib/assets/javascripts/_modules/search.js
```diff
@@ -65,7 +67,7 @@
       // We need JavaScript to do search, so if JS is not available the search
       // input sends the query string to Google. This JS function changes the
       // input to instead send it to the search page.
-      $searchForm.prop('action', '/search')
+      $searchForm.prop('action', pathToSiteRoot + 'search/index.html')
```
lfdebrux, line 70: I think it's best if this doesn't have the `index.html` file name, to be consistent with other URLs. Is there a reason that you changed it?

```suggestion
      $searchForm.prop('action', pathToSiteRoot + 'search/')
```

eddgrant: If we don't include the `index.html` then I _think_ we become reliant on the web server that the site is deployed behind being configured to have a default page value of `index.html`. Having said that I think there are other areas where we generate links which rely on the default page behaviour, so you could argue this is incongruous with that approach. Happy to try taking it out if you'd prefer consistency here.

lib/govuk_tech_docs/path_helpers.rb
eddgrant: Could you give an example? I'm not quite sure what behaviour we'd be after here.

lfdebrux: I think, [based on what the `middleman-search` library does](https://github.com/manastech/middleman-search/blob/ed9e5254bb554988c755fc40238d9e9f67b20abe/lib/middleman-search/extension.rb#L29), that we just need to have `path_to_site_root` be `config[:http_prefix]`:

```suggestion
        path_to_site_root = config[:http_prefix]
```

I think `:http_prefix` should always be the path to the site root, even if the user hasn't defined it, and it should already default to `/`:

https://github.com/middleman/middleman/blob/97ddbc52eac7a937adcc7d418fa9c2fb4a0a1a1e/middleman-core/lib/middleman-core/application.rb#L154

eddgrant: Have just pushed a change to incorporate this.

lib/assets/javascripts/_modules/search.js
```diff
@@ -149,7 +151,8 @@
         var content = s.processContent(result.content, query)
         output += '<li class="search-result">'
         output += '<h3 class="search-result__title">'
-        output += '<a href="' + result.url + '">'
+        var url = result.url.startsWith('/') ? result.url.slice(1) : result.url
+        output += '<a href="' + pathToSiteRoot + url + '">'
```
lfdebrux, line 155: Ah, sorry, my previous suggestion broke things slightly 😅 

It turns out that the search index in `search.json` includes `http_prefix` in the URLs, so when using http prefix and not relative links we now end up with the http prefix duplicated 🤔 .

lfdebrux: I think the easiest way to fix this is to check if `pathToSiteRoot` is a relative link, and only append if it is

```suggestion
        url = pathToSiteRoot.startsWith('..') ? pathToSiteRoot + url : '/' + url  // url already contains pathToSiteRoot when using :http_prefix
        output += '<a href="' + url + '">'
```

There might be a neater way to do that though 😅

eddgrant: Aah, good spot @lfdebrux ! Will take a look. Out of interest, have you been able to get a site running locally when specifying `http_prefix`? All the sites I've tried break horribly as none of the assets are available under their prefixed URLs :thinking: .

eddgrant: hmm, I'm a little confused by this. I _think_ it has been caused by this change in `path_helpers.rb`:

```ruby
path_to_site_root = config[:http_prefix]
```

After the change, this results in any configured `http_prefix` being added in the path to the site root, which is then fed down in to the search system, breaking the search.

[This (rather historical) comment](https://github.com/middleman/middleman/issues/484#issuecomment-6459631) seems to suggest that `http_prefix` is only intended for use to prefix assets (for example when using a CDN and needing to specify some prefix). Having read that I'm now feeling that our use of it to calculate the site root is contrary to its intended use. Unfortunately the issue that comment comes from was never resolved, despite nearly a year of discussion :disappointed:.

I think my preference would be to revert back to reporting the site root as `/` in `path_helpers.rb` when using absolute links, this feels more in keeping with the intent of `http_prefix` to me. Grateful for your thoughts @lfdebrux .

lfdebrux: > Out of interest, have you been able to get a site running locally when specifying http_prefix? All the sites I've tried break horribly as none of the assets are available under their prefixed URLs 🤔 .

Yes, the assets do cause issues, however we recently closed one PR to fix that partially #197, and I have a PR open to fix that for govuk-frontend assets #292. There's also a fix needed for the path `search.json`, in #196, although I'm hoping that we can fix that for both http_prefix and relative_links in this PR 🤞. 

If you're interested in how I've configured everything, right now I have a couple of copies of the example site building together, with different configs, and I serve them from one directory using a Python test server.

In short what I'm doing is:

```
cd example
bundle exec middleman build --http-prefix=/with-http-prefix/
cd ../..
mkdir public
cp -r tech-docs-gem/example/build public/with-http-prefix
cd public
python3 -m http.server
```

That seems to work for me, if I have the right fixes in place. I think copying (or symbolic linking) the build into a folder with the same name as the http_prefix is important; it might also be important to make sure the http_prefix begins and ends with a slash, although I haven't played around with that too much.

eddgrant: Sorry for the delay in replying @lfdebrux , thanks for the tips on getting things running with an `http_prefix`, that was really helpful. I have just pushed what I hope is an appropriate fix to ensure that `http_prefix` is no longer duplicated when we generate the links to the search result pages. I have tested it in all the use cases I can think of (relative links, absolute links with and without an http_prefix) but would be grateful if you could give it some :eyes: too. Cheers!

lfdebrux: Awesome, thanks! I'll take a look 👀

eddgrant:
I've just spotted the conflict arising between 3c629e753b480b85c4618ec850c32a42286c2e5a (recently merged to master) and 3e8770e3ad45ddeefceae688d6da5f84cae596e1 (my change). I'm not 100% sure yet but I think the change merged to master might be a more idiomatic as it seems to use a variable which is exposed by the `middleman-search` gem, rather than relying on my method of walking back to the site root.

I will do some testing to see if we can replace my change with the one already merged.

eddgrant:
After inspection I can now see that the 2 commits do subtly different things. One provides the relative path to the site root and the other provides the path to the search index page. I have rebased my changes to include the other change and everything seems to be working.

If I push my rebased branch is it going to destroy all of our conversation above @lfdebrux ? :thinking: If it will (and that's a problem) then I could perhaps try pulling the changes in as a merge commit instead. Let me know what you'd rather?

lfdebrux:
@eddgrant thanks for looking into it! A rebase and force push won't affect the conversation on this pull request page, so feel free to go ahead. It's preferred to a merge commit anyway.

lib/assets/javascripts/_modules/search.js
```diff
@@ -51,7 +53,7 @@
     this.downloadSearchIndex = function downloadSearchIndex () {
       updateTitle('Loading search results')
       $.ajax({
-        url: searchIndexPath,
+        url: pathToSiteRoot + (searchIndexPath.startsWith('/') ? searchIndexPath.slice(1) : searchIndexPath),
```
lfdebrux, line 56: Unfortunately this means we're getting `http_prefix/http_prefix` here as well.

I'd suggest we remove some of the complexity here and go back to just hardcoding `search.json`. Then we don't need both `data-search-index-path` and `data-path-to-site-root`.

```suggestion
        url: pathToSiteRoot + (searchIndexPath.startsWith('/') ? searchIndexPath.slice(1) : searchIndexPath),
```

lib/source/layouts/_search.erb
```diff
@@ -1,5 +1,5 @@
 <% if config[:tech_docs][:enable_search] %>
-<div class="search" data-module="search" data-search-index-path="<%= search_index_path %>">
+<div class="search" data-module="search" data-path-to-site-root="<%= path_to_site_root config, current_page.path %>" data-search-index-path="<%= search_index_path %>">
```
lfdebrux, line 2:
```suggestion
<div class="search" data-module="search" data-path-to-site-root="<%= path_to_site_root config, current_page.path %>">
```

lfdebrux (commented):
I think this is almost ready to be merged <3, just one small suggested change 🙏

lib/assets/javascripts/_modules/search.js
eddgrant: Aah that's a shame. Happy to revert to a hardcoded `search.json`, however I should note we'll be undoing the changes made in https://github.com/alphagov/tech-docs-gem/commit/3c629e753b480b85c4618ec850c32a42286c2e5a. Is that ok?

lfdebrux: Yep, fine by me.

eddgrant:
@lfdebrux sorry for the delay, it's been a busy week. I've just pushed the change reverting back to always using `/search.json` for the search results.

lfdebrux:
> @lfdebrux sorry for the delay, it's been a busy week. I've just pushed the change reverting back to always using `/search.json` for the search results.

No worries, thanks so much for all the time you've spent on this!

lfdebrux (approved):

eddgrant:
No worries at all, likewise thanks for all your effort reviewing it! :pray:

lfdebrux:
> Hey @lfdebrux I forgot to ask, do you have plans to tag this as a release and publish a new rubygem? I'm interested so I can plan when to deprecate our fork and move back to using this directly.

@eddgrant I've discussed this with the other working group members, we will try and put out a release soon.

lfdebrux:
This issue has been fixed in [release v3.2.0](https://github.com/alphagov/tech-docs-gem/releases/tag/v3.2.0) 🎉 

You can now [configure your Tech Docs Template (TDT) to build your documentation site to use relative links to pages and assets](https://tdt-documentation.london.cloudapps.digital/configure_project/global_configuration/#build-your-site-using-relative-links).

Thanks [@eddgrant](https://github.com/eddgrant) for contributing this feature and the associated fixes.

This change was introduced in [pull request #291: Support sites deployed on paths other than "/" (by generating relative links)](https://github.com/alphagov/tech-docs-gem/pull/291).

PR #292, opened 2022-01-25
Make GOV.UK Frontend assets paths work with http prefix

lfdebrux:
## What’s changed

You no longer need to change the Sass variable `govuk-assets-path` when using a http prefix; the tech docs gem will add the prefix for you.

## Identifying a user need

Users who want to deploy their tech docs to GitHub pages need to handle an additional path prefix to all URLs; they may want to use the `http_prefix` setting for this. PR #197 added this http prefix to all assets from the tech docs gem, but required the user to also override `govuk-assets-path` themselves so that assets from GOV.UK Frontend had the correct path. This PR should mean that this additional step is no longer required, making configuration easier for users.

### DIFF
--- a/lib/assets/stylesheets/_govuk_tech_docs.scss
+++ b/lib/assets/stylesheets/_govuk_tech_docs.scss
@@ -1,4 +1,20 @@
-$govuk-assets-path: "/assets/govuk/assets/" !default;
+// Set $govuk-assets-path so that $govuk-images-path and $govuk-fonts-path
+// (used below, but set in govuk/base) are correct.
+$govuk-assets-path: "/assets/govuk/assets/";
+
+// Use Sprockets' asset url helpers to add the http prefix if present.
+// GOV.UK Frontend provides hooks for us to define the url lookup functions,
+// see https://frontend.design-system.service.gov.uk/sass-api-reference/
+@function tech-docs-image-url-function($filename) {
+  @return image-url($govuk-images-path + $filename);
+}
+
+@function tech-docs-font-url-function($filename) {
+  @return font-url($govuk-fonts-path + $filename);
+}
+
+$govuk-font-url-function: "tech-docs-font-url-function";
+$govuk-image-url-function: "tech-docs-image-url-function";
 
 // Include only the bits of GOV.UK Frontend we need
 $govuk-new-link-styles: true;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Fixes
+
+You no longer need to change the Sass variable govuk-assets-path when using a http prefix; the tech docs gem will add the prefix for you.
+
+- [#292: Make GOV.UK Frontend assets path work with http prefix](https://github.com/alphagov/tech-docs-gem/pull/292)
+
 ## 3.2.1
 
 ### Fixes
--- a/lib/govuk_tech_docs.rb
+++ b/lib/govuk_tech_docs.rb
@@ -24,12 +24,24 @@ require "govuk_tech_docs/warning_text_extension"
 require "govuk_tech_docs/api_reference/api_reference_extension"
 
 module GovukTechDocs
+  class GovukTechDocsExtension < Middleman::Extension
+    def after_configuration
+      if app.config[:govuk_assets_path]
+        warn "Warning: setting govuk_assets_path in config.rb is no longer necessary, and will have no effect."
+      end
+    end
+  end
+
+  ::Middleman::Extensions.register(:govuk_tech_docs, GovukTechDocsExtension)
+
   # Configure the tech docs template
   #
   # @param options [Hash]
   # @option options [Hash] livereload Options to pass to the `livereload`
   #   extension. Hash with symbols as keys.
   def self.configure(context, options = {})
+    context.activate :govuk_tech_docs
+
     context.activate :sprockets
 
     context.sprockets.append_path File.join(__dir__, "../node_modules/govuk-frontend/")


36degrees (commented):
If I understand what this is doing correctly, then I think there's a neater way to do this that doesn't need us to monkey patch the `asset_path` function.

GOV.UK Frontend allows us to define which function to use when generating paths for images ([`$govuk-image-url-function`](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-image-url-function)) and the font ([`$govuk-font-url-function`](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-font-url-function)).

If no custom function is set, the default behaviour is just to prefix the `$filename` with `$govuk-images-path` / `$govuk-fonts-path` (which default to `"#{$govuk-assets-path}images/"` and `"#{$govuk-assets-path}fonts/"` respectively).

I think we can instead pass our own function instead which does the same thing but then calls `asset-path` with the resolved path _including the filename_. So something like this:

```scss
$govuk-assets-path: "/assets/govuk/assets/";

@function tech-docs-image-handler($filename) {
  @return asset-path($govuk-image-url-function + $filename);
}

$govuk-image-url-function: 'tech-docs-image-handler';
```

As ever, happy to talk through it if it's helpful!

lfdebrux:
> If I understand what this is doing correctly, then I think there's a neater way to do this that doesn't need us to monkey patch the `asset_path` function.
> 
> GOV.UK Frontend allows us to define which function to use when generating paths for images ([`$govuk-image-url-function`](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-image-url-function)) and the font ([`$govuk-font-url-function`](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-font-url-function)).
> 
> If no custom function is set, the default behaviour is just to prefix the `$filename` with `$govuk-images-path` / `$govuk-fonts-path` (which default to `"#{$govuk-assets-path}images/"` and `"#{$govuk-assets-path}fonts/"` respectively).
> 
> I think we can instead pass our own function instead which does the same thing but then calls `asset-path` with the resolved path _including the filename_. So something like this:
> 
> ```css-scss
> $govuk-assets-path: "/assets/govuk/assets/";
> 
> @function tech-docs-image-handler($filename) {
>   @return asset-path($govuk-image-url-function + $filename);
> }
> 
> $govuk-image-url-function: 'tech-docs-image-handler';
> ```
> 
> As ever, happy to talk through it if it's helpful!

Thanks for this @36degrees, this is much nicer!

I've pushed a different version of the code now based on your suggestion.

lib/assets/stylesheets/_govuk_tech_docs.scss
```diff
@@ -1,4 +1,15 @@
-$govuk-assets-path: "/assets/govuk/assets/" !default;
+@function tech-docs-image-url-function($filename) {
+  @return image-url($govuk-images-path + $filename)
+}
+
+@function tech-docs-font-url-function($filename) {
+  @return font-url($govuk-fonts-path + $filename)
+}
```
36degrees, line 7: Minor but it looks like these need semicolons for the syntax highlighting to work correctly

```suggestion
@function tech-docs-image-url-function($filename) {
  @return image-url($govuk-images-path + $filename);
}

@function tech-docs-font-url-function($filename) {
  @return font-url($govuk-fonts-path + $filename);
}
```

lib/assets/stylesheets/_govuk_tech_docs.scss
```diff
@@ -1,4 +1,15 @@
...
+
+$govuk-font-url-function: 'tech-docs-font-url-function';
+$govuk-image-url-function: 'tech-docs-image-url-function';
+
+$govuk-assets-path: "/assets/govuk/assets/";
```
36degrees, line 12: I suspect this does work as intended, because presumably `$govuk-fonts-path` isn't evaluated until the function is called, but it might make people's heads hurt less if this variable was set before the functions were defined

(I appreciate that `$govuk-images-path` and `$govuk-fonts-path` haven't been defined at all at this point!)

36degrees (commented):
I'm guessing there's no easy way to test this, given the functionality we're interested in is really part of Sprockets?

Otherwise I'm happy with this approach, bar a couple of very minor comments.

lib/assets/stylesheets/_govuk_tech_docs.scss
36degrees: Weird how the syntax highlighting works correctly in the diff 🤷🏻

lfdebrux:
> I'm guessing there's no easy way to test this, given the functionality we're interested in is really part of Sprockets?
> 
> Otherwise I'm happy with this approach, bar a couple of very minor comments.

Ah tests is a good idea, I'll add some.

lib/assets/stylesheets/_govuk_tech_docs.scss
lfdebrux: Huh, weird. Thanks for catching that, I didn't realise Sass is semicolon-y.

lib/assets/stylesheets/_govuk_tech_docs.scss
lfdebrux: Hmm, yes, I don't know, I kind of thought that having it after would help people realise that `$govuk-{images,fonts}-path` are not defined here and that they need to dig deeper to understand it better. Maybe a comment would be more kind though 😂

lfdebrux: I've added some comments, let me know if you think more detail is needed.

JC-DSIT:
Gonna close this here as it's quite old and needs a review in with the latest changes.  Have added it to issue #489 for tracking

PR #321, opened 2023-02-17
Add api_operation to only render docs for a single operation on an endpoint

ethanmills:
## What’s changed
- Add the option to include a single API operation from openapi spec
- Document existing ability to document schema object from openapi spec

## Identifying a user need
I would like to add some supporting content between operations. Also, `get` and `post` on the same endpoints are used by different types of consumers for some of our endpoints. I would like to be able to split these out into consumer specific pages


ethanmills:
I've also opened https://github.com/alphagov/tdt-documentation/pull/199 for the documentation to go along with this

PR #323, opened 2023-04-11
Fix XSS vulnerability on search results page

ChrisBAshton:
Pages that are indexed in search results have their entire contents indexed, including any HTML code snippets. These HTML snippets would appear in the search results unsanitised, so it was possible to render arbitrary HTML or run arbitrary scripts:

> script being invoked
> HTML being rendered

This is a largely theoretical security issue; to exploit it, an attacker would need to find a way of committing malicious code to a page indexed by a site that uses tech-docs-gem (which are typically not editable by untrusted users). Their code would also be limited by the relatively short length that's rendered in the corresponding search result. Nevertheless, the XSS would then be triggerable by visiting a pre-constructed URL (`/search/index.html?q=some+search+term`), which users could be tricked into clicking on through social engineering.

## What’s changed

This commit sanitises the HTML before rendering it to the page. It does so whilst retaining the `<mark data-markjs="true">` behaviour that highlights the search term in the result:

> sanitised HTML with highlights

I've used jQuery's `text()` function for sanitisation, as that is the approach used elsewhere in the project ([1]).
    
I did consider using native JavaScript (using the same approach as in Mustache [2]) to avoid the jQuery dependency, but this itself may contain bugs and would lead to having two sanitisation approaches to maintain, so I opted against it. For future reference, the code in this commit can be swapped out with:

```js
var entityMap = {
  '&': '&amp;',
  '<': '&lt;',
  '>': '&gt;',
  '"': '&quot;',
  "'": '&#39;',
  '/': '&#x2F;',
  '`': '&#x60;',
  '=': '&#x3D;'
};
var sanitizedContent = String(content).replace(/[&<>"'`=\/]/g, function (s) {
  return entityMap[s];
});
```

[1]: https://github.com/alphagov/tech-docs-gem/blob/66cc7ab0a06dc2f1fe89de8cba2270fcf46f6466/lib/assets/javascripts/_modules/search.js#L202-L204
[2]: https://github.com/janl/mustache.js/blob/972fd2b27a036888acfcb60d6119317744fac7ee/mustache.js#L60-L75

## Identifying a user need

The look and interactions of the gem are unchanged. This simply addresses a security issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+This change solves a potential security issue with HTML snippets. Pages indexed in search results have their entire contents indexed, including any HTML code snippets. These HTML snippets would appear in the search results unsanitised, making it possible to render arbitrary HTML or run arbitrary scripts.
+
+You can see more detail about this issue at [#323: Fix XSS vulnerability on search results page](https://github.com/alphagov/tech-docs-gem/pull/323)
+
 ## 3.3.0
 
 ### New features
--- a/lib/assets/javascripts/_modules/search.js
+++ b/lib/assets/javascripts/_modules/search.js
@@ -169,8 +169,8 @@
 
     this.processContent = function processContent (content, query) {
       var output
-      content = '<div>' + content + '</div>'
-      content = $(content).mark(query)
+      var sanitizedContent = $('<div></div>').text(content).html()
+      content = $('<div>' + sanitizedContent + '</div>').mark(query)
 
       // Split content by sentence.
       var sentences = content.html().replace(/(\.+|:|!|\?|\r|\n)("*|'*|\)*|}*|]*)/gm, '|').split('|')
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- [#323: Fix XSS vulnerability on search results page](https://github.com/alphagov/tech-docs-gem/pull/323)
```
claireashworth, line 5:
```suggestion
This change solves a potential security issue with HTML snippets. Pages indexed in search results have their entire contents indexed, including any HTML code snippets. These HTML snippets would appear in the search results unsanitised, making it possible to render arbitrary HTML or run arbitrary scripts.

You can see more detail about this issue at [#323: Fix XSS vulnerability on search results page](https://github.com/alphagov/tech-docs-gem/pull/323)
```

claireashworth: I used your issue description to add a bit more detail to the change log.

lib/assets/javascripts/_modules/search.js
```diff
@@ -169,8 +169,9 @@
 
     this.processContent = function processContent (content, query) {
       var output
-      content = '<div>' + content + '</div>'
-      content = $(content).mark(query)
+      $('html').append('<div id="tmp-div-for-sanitising-html"></div>')
+      var sanitizedContent = $('#tmp-div-for-sanitising-html').detach().text(content).html()
```
lfdebrux, line 173: This worked for me, without the need to attach and then detach a node:

```suggestion
      var sanitizedContent = $('<div id="tmp-div-for-sanitising-html"></div>').text(content).html()
```

ChrisBAshton: Ahh, nice one!

lfdebrux (approved):

PR #325, opened 2023-04-14
TRG-668: Support sites deployed on a path other than "/" when generating header and footer links.

eddgrant:
Hey folks :wave: 

## What’s changed

[v3.2.0](https://github.com/alphagov/tech-docs-gem/releases/tag/v3.2.0) contained changes to support support sites which were configured to generated relative links. At the time I implemented these changes we were not making use of the `header_links` and `footer_links` links functionality. We have since started using both which caused me to realise that the relative link generation did not work for either.

This PR extends the relative link generation functionality such that it works for both header and footer links. I have also updated the tests to be more comprehensive so that they test link generation for both internal and external links, as I found that external links were not being correctly identified in all cases previously.

Hope this all makes sense. Would be great if someone could give it some :eyes: and consider it for inclusion. Very happy to discuss and fettle if required.

Cheers!


lib/govuk_tech_docs/path_helpers.rb
```diff
@@ -26,5 +38,25 @@ def path_to_site_root(config, page_path)
...
+    def get_resource_path_relative_to_current_page(config, current_page, resource_path_from_site_root)
+      path_segments = resource_path_from_site_root.split("/").reject(&:empty?)[0..-2]
+      path_file_name = resource_path_from_site_root.split("/")[-1]
+
+      path_to_site_root = path_to_site_root config, current_page
+      path_to_site_root + path_segments.push(path_file_name).join("/")
+    end
+
+    def external_url?(url)
+      puts "URL is #{url}"
```
lfdebrux, line 57: Was this included just for debugging? If so I think it should be removed now

lib/govuk_tech_docs/path_helpers.rb
```diff
@@ -1,18 +1,30 @@
 module GovukTechDocs
   module PathHelpers
-    def get_path_to_resource(config, resource, current_page)
-      if config[:relative_links]
-        resource_path_segments = resource.path.split("/").reject(&:empty?)[0..-2]
-        resource_file_name = resource.path.split("/")[-1]
+    require "uri"
```
lfdebrux, line 3: Can we put this at the very top of the file?

lib/govuk_tech_docs/path_helpers.rb
eddgrant: Aah good spot. Have fixed.

spec/table_of_contents/helpers_spec.rb
```diff
@@ -129,6 +129,10 @@ def render(_layout)
           def add_children(children)
             @children.concat children
           end
+
+          def is_a?(clazz)
+            clazz.to_s == "Middleman::Sitemap::Resource"
```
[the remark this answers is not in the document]
eddgrant: D'oh, my years of JVM programming coming back to haunt me :laughing: . Have changed this now.

lfdebrux (approved):

Issue #318, opened 2023-01-04
Error: Filters is not a module


lfdebrux:
Some users have reported seeing this error message after upgrading dependencies:

```
bundler: failed to load command: middleman (.bundle/bin/middleman)
TypeError: Filters is not a module
  .bundle/gems/middleman-syntax-3.2.0/lib/middleman-syntax/haml_monkey_patch.rb:4:in `<module:Haml>'
```

It seems to be related to using HAML >= 6, downgrading HAML to ~> 5 works as a workaround.

lfdebrux:
Fixing this in #328.

PR #328, opened 2023-05-19
Add tests for Ruby 3.x support

### LINKED ISSUE #327: Update to support Ruby v3
## What should change

The Gem and dependencies should be updated to support Ruby v3

## User need

Ruby v2 has been end-of-life since 31-Mar-23 https://endoflife.date/ruby

lfdebrux:
## What’s changed

Ruby 2.7 has reached end of life, our users want to be able to use the latest stable versions of Ruby. Middleman recently released support for Ruby 3.1 and greater [[1]], so this gem should be able to work with Ruby 3, but to make sure we should update our test workflows. A few tweaks were needed, including updating the linter, but nothing major.

At the same time we also update the version of Node.js used for the tests, as Node.js 14 reached end of life a while ago as well. I also threw in a fix for the Haml issues (#318), as the tests wouldn't run otherwise.

[1]: https://github.com/middleman/middleman/pull/2635

## Identifying a user need

Closes #327.


huwd (commented):
Looks good, and CI seems happy.
Great to have node upgraded too.

we're removing the .ruby-version file, interested in the thinking there?
Whilst there's no Gemfile here with the usual referencing of that file, the ruby-version does have some general use, no?

Don't both Rbenv and RVM use that file to try and help setup a devs ruby version?

Might be a gap in my knowledge as I mostly work on applications not gems 🤔

huwd (approved):
Ace, seems like a sensible reason to me.
✅ 

I guess if anyone really misses it and it justifies the maintenance they can add it back in later with a commit that says so.

jfharden:
Removing the .ruby-version file has broken the publishing to rubygems step: https://github.com/alphagov/tech-docs-gem/actions/runs/5023630866/jobs/9008447497

jfharden:
I've created a new issue for it: https://github.com/alphagov/tech-docs-gem/issues/329

lfdebrux:
Ah, good spot @jfharden, thanks for flagging. I hadn't noticed as I hadn't been trying to publish these changes.

PR #335, opened 2023-11-01
Calculate redirects as relative paths when site is configured to use relative links.

eddgrant:
## What’s changed

[v3.2.0](https://github.com/alphagov/tech-docs-gem/releases/tag/v3.2.0) contained changes to support support sites which were configured to generated relative links. At the time I implemented these changes we were not making use of the `old_paths` page redirection functionality. We have since started using this which caused me to realise that the redirect page generation did not work correctly when a site is configured to use relative links.

This manifested in the redirect pages being create in the wrong location in the site hierarchy and their redirect URLs being wrong (pointing to paths which don't exist).

This PR updates the redirect functionality such that it now works correctly when sites are configured to use relative links . I have written some simple redirect unit tests as non existed previously.

Hope this all makes sense. Would be great if someone could give it some 👀 and consider it for inclusion. Very happy to discuss and fettle if required.

Cheers!

## Identifying a user need

No UR done. User testing was what actually alerted us to this issue in the first place though, as users were not able to access redirect pages, because they were being generated incorrectly.

User story:

```
As an author of a tech-docs-gem based site which uses relative links
I want to be able to provide my users with redirect pages
So that I can confidently rename and relocate my pages, without degrading the experience of my users
```

### DIFF
--- a/lib/govuk_tech_docs/path_helpers.rb
+++ b/lib/govuk_tech_docs/path_helpers.rb
@@ -1,26 +1,28 @@
 require "uri"
 module GovukTechDocs
   module PathHelpers
-    # Some useful notes from https://www.rubydoc.info/github/middleman/middleman/Middleman/Sitemap/Resource#url-instance_method :
-    # 'resource.path' is "The source path of this resource (relative to the source directory, without template extensions)."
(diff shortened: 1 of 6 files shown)


JC-DSIT:
As this is very old I'm closing it and linking it to issue #489 for review

PR #338, opened 2024-01-26
Update rubocop-govuk to v4.13.0

huwd:
## What’s changed

Depends on: https://github.com/alphagov/tech-docs-gem/pull/339
Bump rubocop-govuk to v4.13.0

## Identifying a user need

Dependency updates need to happen otherwise folks will start forking this repo and it's relevance declines.


sengi:
Happened as part of #349.

PR #339, opened 2024-01-26
Shift required_ruby_version to v3.1.4

huwd:
## What’s changed

Minimum supported ruby is now 3.1.4

Breaking change:
2.7x was end of support 31 Mar 2023
We'll run into dependency issues (expectations having moved up) if we don't upgrade.

Let's move the testing to target the 3x series, but enforce 3 as a new minimum.

## Identifying a user need


sengi:
Happened as part of #349.

PR #340, opened 2024-01-26
Update middleman

huwd:
## Please fill in the sections below

Bump middleman related gems up, all minor changes

## Identifying a user need


sengi:
Middleman updated as far as [currently compatible](https://github.com/alphagov/tech-docs-gem/commit/1764967) in #349.

PR #341, opened 2024-01-26
Update Rspec to v3.12.0 & Jasmine to v3.99.0

huwd:
## What’s changed

Update rspec, minor version bump only

## Identifying a user need

### DIFF
--- a/govuk_tech_docs.gemspec
+++ b/govuk_tech_docs.gemspec
@@ -51,7 +51,7 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency "byebug"
   spec.add_development_dependency "capybara", "~> 3.32"
-  spec.add_development_dependency "jasmine", "~> 3.5.0"
+  spec.add_development_dependency "jasmine", "~> 3.99.0"
   spec.add_development_dependency "rake", "~> 13.0"
   spec.add_development_dependency "rspec", "~> 3.12.0"
   spec.add_development_dependency "rubocop-govuk", "~> 4.10.0"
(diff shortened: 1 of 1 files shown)


huwd:
Note Jasmine gem is End of Life.

```
The Jasmine Ruby gems are deprecated. There will be no further releases after                                                                                                                 
the end of the Jasmine 3.x series. We recommend that most users migrate to the                                                                                                                
jasmine-browser-runner npm package, which is the direct replacement for the                                                                                                                   
jasmine gem. See <https://jasmine.github.io/setup/browser.html> for setup                                                                                                                     
instructions, including for Rails applications that use either Sprockets or                                                                                                                   
Webpacker.                                                                                                                                                                                    
                                                                                                                                                                                              
If jasmine-browser-runner doesn't meet your needs, one of these might:                                                                                                                        
                                                                                                                                                                                              
* The jasmine npm package to run specs in Node.js:                                                                                                                                            
  <https://github.com/jasmine/jasmine-npm>                                                                                                                                                    
* The standalone distribution to run specs in browsers with no additional                                                                                                                     
  tools: <https://github.com/jasmine/jasmine#installation>                                                                                                                                    
* The jasmine-core npm package if all you need is the Jasmine assets:                                                                                                                         
  <https://github.com/jasmine/jasmine>. This is the direct equivalent of the                                                                                                                  
  jasmine-core Ruby gem.                                                                                                                                                                      
```

We should update this along with the readme. Separate PR however

sengi:
Happened as part of #349 (and version constraints loosened to pin just the major, so should be less update toil in future).

PR #350, opened 2024-05-09
Support for GA4 Tag Manager

heathd:
adding support for GA4 tag manager. Can activate by setting `ga4_tracking_id` config variable.

I have not reviewed the custom analytics tracking javascript relating to search operations, so that may not work

## What’s changed

Add support for GA4 tag manager

## Identifying a user need

The GDS Engineering Enablement team need this as we are using TDT for our team documentation and would like to add google analytics support

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,7 +67,7 @@ To use the service_link at the start of the path to the favicon images, you need
 
 - [Allow non-GOV.UK favicon and opengraph assets](https://github.com/alphagov/tech-docs-gem/pull/387)
 
-To use a non-crown assets, you need to 
+To use a non-crown assets, you need to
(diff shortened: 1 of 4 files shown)


huwd:
Let's look at this for 4.1 there's a GTM PR below too would like to consider the two at the same time

davidread:
@huwd Yes please, could you review and merge this? We're currently maintaining a fork until this is done.

[Google Universal Analytics is dead](https://support.google.com/analytics/answer/11583528?hl=en), which drives the need for supporting GA4, to avoid going backwards

huwd:
Yeah looks reasonable and additive.

huwd:
Interestingly fails on `main` now too:
https://github.com/alphagov/tech-docs-gem/actions/runs/16609419271/job/46989177300

Issue #357, opened 2024-09-06
Table of contents section has role=dialog on desktop


tombye:
## What should change

The table of contents shouldn't have role=dialog when it doesn't have dialog behaviours (on screens larger than mobile size).

## User need

On mobile, the table of contents section acts like a [dialog](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role) because It:
- is hidden by default
- is shown when you activate a button
- is focused when shown
- has an accessible name
- constrains tabbing within the dialog

It has a role=dialog, which makes sense there.

<img width="318" alt="image" src="https://github.com/user-attachments/assets/78fb2980-3a55-4293-8deb-6e942044d198">

On desktop, it doesn't behave like this so shouldn't have that role:

<img width="1727" alt="image" src="https://github.com/user-attachments/assets/67b7e4fd-8997-438b-b7cf-76599c944713">

This effectively leaves it with the wrong semantics which, I think, would make it fail [success criterian 1.3.1 - info and relationships](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html).

tombye:
This has been fixed by https://github.com/alphagov/tech-docs-gem/pull/358 so closing.

PR #358, opened 2024-09-12
Only use dialog on mobile

tombye:
## What’s changed

Changed the role of the table of contents so it's only a dialog on mobile-sized screens, to better match its behaviour.

A proposed fix for https://github.com/alphagov/tech-docs-gem/issues/357

The table of contents JS didn't have tests so I added some before making these changes.

## Identifying a user need

Users need parts of the page to have roles that match their behaviour, otherwise it's confusing because it doesn't match their expectations. This is most obvious when you're using a screen reader and the tech docs are displayed on a desktop-size screen. The table of contents is announced as a dialog but it doesn't popup when you click a button like a dialog or have any of the other associated behaviours like tabbing being constrained to the dialog.

I chose to remove the role completely on desktop because its contents are already wrapped in landmark roles ('search' and 'navigation') and every landmark you add introduces more stuff to navigate.


kr8n3r (approved):
🎉

PR #359, opened 2024-09-13
Bump ruby version for the example project

marcotranchino:
## What’s changed

With this commit, we are just updating the ruby version for the example project since the current version is old and no longer supported.

## Identifying a user need

Prevent users from having to download an old unsupported ruby version.

## Screenshot of the example project

Working fine with the new ruby version.

example project with ruby 3 3 0


huwd (approved):
great let's merge it

Issue #356, opened 2024-09-06
Focused links can be covered by the table of contents button on mobile


tombye:
## What should change

The table of contents button should be changed for smaller screens so it can't overlap links when they are focused.

## User need

If you rely on the keyboard, you need to see what's in focus at all times. If a link gets hidden, it also breaks [success criterion 2.4.11, Focus not obscured (minimum)](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum.html) and [success criterion 2.4.12, Focus not obscured (enhanced)](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-enhanced.html).

For example, see the link to the 'GOV.UK kubernetes manual' on [https://docs.publishing.service.gov.uk/](https://docs.publishing.service.gov.uk/manual/architecture-deep-dive.html) receiving focus while hidden:

![mobile_tech_docs_template_focus_obscured_fail](https://github.com/user-attachments/assets/7bb09015-4f0d-473c-bd63-773a3af193ab)

tombye:
We added a fix for this in https://github.com/alphagov/tech-docs-gem/pull/360 so I'm marking this as closed.

PR #360, opened 2024-09-17
Stop table of contents sticky header overlapping focused items

tombye:
Stop the table of contents sticky header, which appears on smaller screens, from overlapping interactice elements when they are focused.

A fix for https://github.com/alphagov/tech-docs-gem/issues/356, using a variant of the approach in https://www.tpgi.com/prevent-focused-elements-from-being-obscured-by-sticky-headers/.

### Before

mobile_tech_docs_template_focus_obscured_fail

Tabbing backwards through content to a link obscured by the table of contents header means the focus indicator is lost.

### After

mobile_tech_docs_template_focus_obscured_pass

Tabbing backwards through content to a link obscured by the table of contents header reveals the link when it it focused.

### Notes for reviewers

This work is done on top of https://github.com/alphagov/tech-docs-gem/pull/358, because that adds tests for the table of contents JS that this adds to. So I don't need to duplicate the work adding them.

## Identifying a user need

Users who use the keyboard to move through the page rely on being able to see what is focused to know where they are in the page.

The issue also failed [success criterion 2.4.11, Focus not obscured (minimum)](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum.html) and [success criterion 2.4.12, Focus not obscured (enhanced)](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-enhanced.html).


kr8n3r (approved):
works as expected 🎉

Issue #352, opened 2024-09-05
Upgrading to govuk-frontend v5


kr8n3r:
As part of maintenance and WCAG 2.2 compliance, we'd like to use the latest major version of [govuk-frontend](https://github.com/alphagov/govuk-frontend/releases/tag/v5.6.0)
Version 5 Javascript support is limited to this support grades https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/browser-support.md  as they use ES modules
"Only browsers in grades A, B and C will run our JavaScript enhancements. We will not support our JavaScript enhancements for older browsers in grade X"

We need to decide if 
a) we're going to use babel to transpile to ES5 or
b) build govuk-frontend Javascript as intended and serve a no-js experience for browsers that don't support it. 

Most of the Javascript modules here are ES5 using global `window` so we'll need to think about potentially rewriting those to ES modules.

Additionally, downstream projects can write and include their own additions to application.js. We can expect that their javascript is likely to be ES5 and possibly using jQuery as well.

One of the options is to generate two files: ESM and non-ESM and bundle downstream into the latter. [Similar to what we're doing on notify-admin for now](https://github.com/alphagov/notifications-admin/blob/main/gulpfile.js#L66-L140)

huwd:
A concern I have is us changing the asset pipeline in a way which impacts any JS currently written in various tech docs sites around.

If we needed for force a rewrite there is dramatically raises the cost of this as a breaking change.
Any thoughts there on if we could bundle both ESM and non-ESM and then ensure that any existing JS in the doc site remains compatible during build?

If we did that then this might not even be a breaking change?

kr8n3r:
> A concern I have is us changing the asset pipeline in a way which impacts any JS currently written in various tech docs sites around.
> 
> If we needed for force a rewrite there is dramatically raises the cost of this as a breaking change. Any thoughts there on if we could bundle both ESM and non-ESM and then ensure that any existing JS in the doc site remains compatible during build?
> 
> If we did that then this might not even be a breaking change?

yes do "generate two files: ESM and non-ESM and bundle downstream into the latter"

then anyone who currently adds their own javascript into the `application.js` like https://github.com/alphagov/govwifi-tech-docs/blob/master/source/javascripts/application.js would be able to do so.
if we need to use any govuk-frontend JS in the gem, then we'd create an ESM bundle

huwd:
oops forgot to push this earlier

I've started a branch here:
[...]

PR #361, opened 2024-09-17
Use `terser` for Javascript minification

marcotranchino:
This is needed because of the need to create two different JS rollups and because the default compressor fails. See link below.

https://github.com/alphagov/tech-docs-gem/issues/352#issuecomment-2352404262


kr8n3r:
i'll close this as distributed dependencies are done differently. Minifier works on my brach

PR #363, opened 2024-09-17
Upgrade to use govuk-frontend v5

kr8n3r:
## What’s changed

- updated asset pipeline and packages used to be able to import govuk-frontend v5
- new `govuk_frontend_all` entry point that imports `all.bundle` from govuk-frontend
- updated javascript and sass compilers
- updated location paths for govuk-frontend
- updated markup as per govuk-frontend release notes

## Identifying a user need

These changes serve two purposes:
1. bring tech-docs frontend dependencies up to date from a maintenance / upgradability point of view
2. bring with it some WCAG 2.2 compliance govuk-frontend offers so that when it comes to assessment / audit, we're in a better position

## How to review / test

See individual commits for more detail
Test in a downstream project:
- clone tech-docs-gem and switch to this branch
- create a `govuk_frontend.js` file in project’s `source/assets/javascripts directory`
- add `//= require govuk_frontend_all` into it
- point to local gem https://github.com/alphagov/tech-docs-gem/tree/use-govuk-frontend-v5?tab=readme-ov-file#use-your-own-tech-docs-template-website
- run `bundle install`
- `run bundle exec middleman server` or `build`


kr8n3r:
jasmine broken. not sure it supports es6

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Changelog
 
 ## Unreleased
+- Upgrade to govuk-frontend v5.6.0 and introduce new Javascript entry point
+
+You need to
+  - create a `govuk_frontend.js` file your project’s `source/assets/javascripts` directory
+  - add `//= require govuk_frontend_all` into it
 
 - BREAKING: drop support for end-of-life Ruby versions 2.7 and 3.0. The minimum Ruby version is now 3.1.
```
huwd, line 10: I think we should drop into the changelog that we've dropped support for ie8

huwd (approved):
Looks great,
@kr8n3r can we consider updating the changelog to signal that we've dropped ie8 in line with designs system upgrades, then let's merge this!

PR #364, opened 2024-09-20
Introduce a custom module to silence dart sass warnings

marcotranchino:
With this change, we have defined a custom SassWarningSupressor module that overrides the warn method so that we can ignore messages containing 'sass' or 'dart-sass' only.

This is needed so that we can avoid sass warnings in downstream projects.


huwd:
What do you make the the test failures @marcotranchino?
Trying to work out if that means phantonJS isn't getting a rendered page in the way we'd want?

chao-xian (approved):

Issue #355, opened 2024-09-06
Images with blank alt text shouldn't be wrapped in anchor tags


tombye:
## What should change

Images with blank alt text shouldn't be wrapped in anchor tags.

## User need

All images are wrapped in anchor tags which link to the image on its own, including images with blank alt text.

When the images appear in-page, we tend to provide the information in them as text in the page. For example:

https://docs.publishing.service.gov.uk/manual/architecture-deep-dive.html

But that still leaves the link with an empty accessible name, because its contents are the image. Based on [this WCAG rule](https://www.w3.org/WAI/standards-guidelines/act/rules/c487ae/), that breaks the following success criteria:
- [2.4.4: Link purpose (in context)](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value)
- [4.1.2: Name, role, value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value)

kr8n3r:
fixed with https://github.com/alphagov/tech-docs-gem/pull/375

PR #366, opened 2024-09-20
Only wrap images with alt text in hyperlinks

tombye:
## What’s changed

This introduces a guard against images with no alt text, choosing to only wrap them in hyperlinks, the default behaviour now, when found.

My Ruby is basic at best but I looked at the method we're overwriting from [the Red Carpet HTML renderer](https://github.com/middleman/middleman/blob/f46306b29443111605dadb8f164ef210393db831/middleman-core/lib/middleman-core/renderers/redcarpet.rb#L74) and copied across its interface more exactly, to make it clear where the `alt_text` variable comes from.

The difference in what users see is minimal. The hyperlinks don't have any special styles, like `display: block` so nothing is different with how they are laid out on the page (both the `<a>` and `` tags default to `display: inline`).

## Identifying a user need

When we wrap images in hyperlinks, we make the image's alt text their only content. If those images have no alt text, they are effectively empty links.

This means they have no accessible name if queried by an accessibility API:

https://w3c.github.io/html-aam/#img-element-accessible-name-computation

...so it'll be up to the screen reader talking to the accessibility API to guess. By testing with the Voiceover, NVDA and JAWs, I screen reders, I found this ends up being the file name, which can't be relied on to explain the image. All in all, this behaviour breaks the non-text content success criterion from WCAG 2.2:

https://www.w3.org/WAI/WCAG22/Understanding/non-text-content.html

This issue is recorded here:

https://github.com/alphagov/tech-docs-gem/issues/355

## Notes for reviewers

Please check my Ruby code! My skills with Ruby are basic at best :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ You need to
 - Use hidden attribute to show/hide expiry notices instead of just CSS
 - Only use dialog role for table of contents when it behaves like one (accessibility fix)
 - Prevent interactive elements being obscured by sticky table of contents header
+- Only wrap images with alt text in hyperlinks
 
 ## 3.5.0
 
--- a/lib/govuk_tech_docs/tech_docs_html_renderer.rb
+++ b/lib/govuk_tech_docs/tech_docs_html_renderer.rb
@@ -20,7 +20,7 @@ module GovukTechDocs
     end
 
     def image(link, title, alt_text)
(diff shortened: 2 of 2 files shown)


tombye:
Thanks @marcotranchino, great suggestions. Now rewritten to include them.

huwd (approved):

PR #368, opened 2024-09-23
Fix intermittent test failures

marcotranchino:
With this change, we are fixing the intermittent test failures by 
updating the tests' setups and afterEach instructions and also by
updating some modules in order to prevent access to properties that 
have not yet been initialised properly.

### DIFF
--- a/lib/assets/javascripts/_modules/table-of-contents.js
+++ b/lib/assets/javascripts/_modules/table-of-contents.js
@@ -10,7 +10,7 @@
     this.isMonitoring = false
   }
   StickyOverlapMonitors.prototype.run = function () {
-    var stickyIsVisible = this.$sticky && this.$sticky.is(':visible')
+    var stickyIsVisible = this.$sticky.length > 0 && this.$sticky.is(':visible')
     if (stickyIsVisible && !this.isMonitoring) {
       document.addEventListener('focus', this.onFocus, true)
       this.isMonitoring = true


tombye:
Thanks for doing this @marcotranchino, the tests failing are a real problem for merging changes at the moment.

Forgive me if I'm wrong but reading the changes to `lib/assets/javascripts/_modules/table-of-contents.js`, I'm not sure they're doing what it looks like. `this.$sticky` is [a jQuery object](https://api.jquery.com/Types/#jQuery), the result of a call to [the $ function](https://api.jquery.com/jQuery/). From what I understand, when called with a CSS selector, [as we're doing here](https://github.com/alphagov/tech-docs-gem/blob/56da6ee47a42f31e61ff2bdf028ed27fca066311/lib/assets/javascripts/_modules/table-of-contents.js#L57), it will always return that object. The only thing that changes is that it will either have DOM nodes, if there are some in the document that match the selector, or not, if there aren't. The normal way to test this is to check the `length` propery.

So I'm not sure how testing the truthyness of `this.$sticky` is making these tests pass because it always returns the object and so should always be truthy. A jQuery object is always passed to the `StickyOverlapMonitors‎` class when its instantiated and so will always be assigned to `this.$sticky`. I guess jQuery might not load but that would cause errors higher up the stack, not where we're seeing them.

I might also just be missing something obvious :) What do you think?

kr8n3r (approved):

PR #374, opened 2024-10-23
Stable sort for left nav

heathd:
The existing sort algorithm for the left menu allows the sort order to be controlled by specifying weights for items however it does not preserve the sort-order of items:
- which have the same weight as one another
- which have no weight assigned

The ruby builtin function `sort_by` *[is not guaranteed to be stable](https://apidock.com/ruby/v2_6_3/Enumerable/sort_by)*. What this means is that if items do not have an explicit weight specified they can randomly shuffle around between each execution of the rendering process which makes it harder than necessary to manage/curate the order of the left menu.

## What’s changed

This PR changes the algorithm to respect the ordering of items as they appear in the middleman resources tree. The algorithm is:

1. items are sorted by weight first
2. items with equal weight are sorted respecting the natural ordering in the resources tree
3. items without a weight specified appear last and are sorted respecting the natural ordering in the resources tree

The overall outcome is that the left nav ordering does not need to be completely controlled by weights. 

This change does not break any existing sites because, if they did not control the ordering entirely by specifying weights, then the ordering would have been non-deterministic.

## User need

This makes it easier to manage the ordering of items in the left nav

### DIFF
--- a/lib/govuk_tech_docs/table_of_contents/helpers.rb
+++ b/lib/govuk_tech_docs/table_of_contents/helpers.rb
@@ -18,12 +18,27 @@ module GovukTechDocs
         output
       end
 
+      # Items with a weight appear first, in weight order
+      #   if items have equal weight, they stay in the order they appeared (stable sort)
+      # Items without a weight appear after, sorted stably in the order that they are
+      # present in the resource list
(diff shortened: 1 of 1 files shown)


kr8n3r:
no changelog?

PR #379, opened 2024-11-14
Disable focus outline on navigation link

kr8n3r:
As we already apply a focus state to the link child element, there is no need for an additional focus outline on the link itself that User Agents apply as default

Before
Screenshot 2024-11-14 at 09 59 15

After
Screenshot 2024-11-14 at 09 59 29


owenatgov (approved):
Sorted

PR #381, opened 2025-01-10
Fix focus issues

tombye:
## What’s changed

Fixes the following accessibility issues discovered when using [Notify's tech docs](https://www.notifications.service.gov.uk/using-notify/api-documentation) with a keyboard:
- when you close the table of contents (toc) dialog, we don't move focus from the close button but do hide it, which is confusing
- the main content pane is focusable on mobile, which doesn't make sense because it's the only scrollable region onscreen (compared to on desktop, where the toc is visible as a separate scrollable region)
- the main content pane is focusbale on desktop but doesn't have a focus style set so you can't see that it is focused when you move to it, particularly when using the skiplink

## Identifying a user need

Something always needs to be in focus when using the keyboard, so you know what you're interacting with. These issues directly affect keyboard users by breaking that in a few places.

### Before

https://github.com/user-attachments/assets/f4e752c7-5884-4d6c-b903-118e05a865f6

### After

https://github.com/user-attachments/assets/95de4d00-d357-4b3f-b000-11ceef8721c3

### Before

https://github.com/user-attachments/assets/0bbfe9a4-0732-445a-97ab-109109b13cff

### After

https://github.com/user-attachments/assets/e7476d6c-d955-4d5b-99a3-8f73c0bfe331

### Before

https://github.com/user-attachments/assets/e18609a0-3b84-4d59-b34c-80b02e94d032

### After

https://github.com/user-attachments/assets/ebe3efd7-d109-4927-b70d-5cf76ddaa700

## Notes for reviewers

Better to read this commit-by-commit.

### DIFF
--- a/lib/assets/javascripts/_modules/table-of-contents.js
+++ b/lib/assets/javascripts/_modules/table-of-contents.js
@@ -113,6 +113,7 @@
       $html.removeClass('toc-open')
 
       toggleBackgroundVisiblity(true)
+      $openButton.focus()
       updateAriaAttributes()
     }
 
--- a/lib/assets/javascripts/_modules/table-of-contents.js
+++ b/lib/assets/javascripts/_modules/table-of-contents.js
@@ -104,7 +104,6 @@
     function openNavigation () {
       $html.addClass('toc-open')
 
-      toggleBackgroundVisiblity(false)
       updateAriaAttributes()
       $toc.focus()
     }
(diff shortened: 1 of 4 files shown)


kr8n3r (approved):
changes look good

kr8n3r:
seems like an issue with latest ubuntu image, ssl version and phantomjs

PR #383, opened 2025-01-15
Release v4.1.2

tombye:
Brings in the following fixes:

- [Fix focus issues with table of contents and main content pane](https://github.com/alphagov/tech-docs-gem/pull/381)

Also includes [these changes to the JS tests](https://github.com/alphagov/tech-docs-gem/pull/382). I haven't included them in the CHANGELOG because they aren't user-facing. If any reviewers think they should be included in the CHANGELOG, please put in a comment and I'll update it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 4.1.2
+
+## Fixes
+
+- [Fix focus issues with table of contents and main content pane](https://github.com/alphagov/tech-docs-gem/pull/381)
+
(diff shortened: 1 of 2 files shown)


huwd (approved):

tombye:
@huwd sorry for the extra work but would you mind checking this again? I adding some tidying up missing from the original changes to jasmine in https://github.com/alphagov/tech-docs-gem/pull/385 so the code this includes has changed.

PR #399, opened 2025-05-01
Use the service_link at the start of the path to the favicon images i…

johnsgp:
## What’s changed

The location of the favicons in the page `<head>` has been changed to include the `service_link` from the [tech-docs.yml](https://github.com/alphagov/tech-docs-gem/blob/main/example/config/tech-docs.yml#L7) configuration file, only if `use_service_link_for_favicon` is set to true (default: false).

## Identifying a user need

The favicon links [added to the core template in v4.0.0](https://github.com/alphagov/tech-docs-gem/commit/01829189c5a9888f532bb06f608035c011e5d109) does not work for sites that have a service link. As an example, see the [Self Assessment End-to-End Service Guide](https://developer.service.hmrc.gov.uk/guides/self-assessment-end-to-end-service-guide/) which has its [source in GitHub](https://github.com/hmrc/self-assessment-end-to-end-service-guide). There is no favicon on the browser tab because the links look like
```<link rel="icon" sizes="48x48" href="/assets/govuk/assets/images/favicon.ico">```
which resolves to https://developer.service.hmrc.gov.uk/assets/govuk/assets/images/favicon.ico which does not exist. The icon is located at https://developer.service.hmrc.gov.uk/guides/self-assessment-end-to-end-service-guide/assets/govuk/assets/images/favicon.ico. The [service_link configuration](https://github.com/hmrc/self-assessment-end-to-end-service-guide/blob/main/config/tech-docs.yml#L5) matches what is missing from the link.

This was raised as an issue by a service guide author. There are [2 dozen or more service guides and roadmaps](https://developer.service.hmrc.gov.uk/api-documentation/docs/api?docTypeFilters=ROADMAPANDSERVICEGUIDE) that either already are or will be affected when updated.

This cannot be applied to all services, so needs an additional switch, `use_service_link_for_favicon` to be set to true.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Unreleased
+
+- [Fix: Use the service link as the start of the path to the favicon images if required](https://github.com/alphagov/tech-docs-gem/pull/399)
+
+To use the service_link at the start of the path to the favicon images, you need to
+- set `use_service_link_for_favicon: true`
+
 ## 4.3.1
 
 - Revert [Fix: Use the service link as the start of the path to the favicon images](https://github.com/alphagov/tech-docs-gem/pull/391)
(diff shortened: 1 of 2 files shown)


kr8n3r:
tested with local docs, works well now

PR #405, opened 2025-06-05
Remove ARIA labels from containers

### LINKED ISSUE #404: Bug: Menu and content containers use aria-label incorrectly
In repositories with a table of contents such as the [WCAG Primer](https://alphagov.github.io/wcag-primer/index.html), both the left hand menu and content window are contained in `<div>` elements which have `aria-label` attributes but no role. This causes inconsistent behaviour across screen readers and browsers.

When viewing on 1280x1021 screen size:

- VoiceOver reads labels out in Chrome and Safari but on Chrome they are being mis-identified as form controls (they do not have any form functionality at this screen size; when text size is increased, page zoomed or mobile view then table of contents is a button).
- JAWS does not read the labels
- NVDA read the "content" label only

Note, if zooming in/out and opening/closing table of contents modal, aria-hidden is set in some scenarios. This state should be consistently maintained so that confusion is not caused.

Code for the menu container: 

```HTML
<div class="toc" data-module="table-of-contents" tabindex="-1" aria-label="Table of contents" aria-hidden="false">
```

Code for the content container:

```HTML
<div class="app-pane__content toc-open-disabled" aria-label="Content" tabindex="0">
```

Strangely this is not detected by Axe (which normally picks these issues up) but it does look like incorrect ARIA usage, as a `<div>` has a default role of `generic` and so `aria-label` is not allowed.

We believe this fails [WCAG 4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value).

This was originally raised as [WCAG Primer issue 133](https://github.com/alphagov/wcag-primer/issues/133).

andrewhick:
## What’s changed

This PR removes unnecessary ARIA labels from the menu container and the main content window.

These aria-labels are used incorrectly as they are applied to `<div>` elements with no role. This is not permitted under the [ARIA specification](https://www.w3.org/TR/wai-aria/#aria-label) as a `<div>` has a role of `generic`. They are also unnecessary as:

- the `<nav>` element already has an appropriate label
- the main content is within a `main` landmark and so an aria-label is unnecessary for its container. We could move the aria-label to `<main>` instead but while permitted, my understanding is that it would be unnecessary.

I've tested it locally using VoiceOver and:

- the side navigation is correctly noted as "Table of contents"
- the `main` landmark is correctly identified as `main`

I believe this fixes #404 .

## Identifying a user need

This came out of testing by Shabana Ali in the WCAG Primer working group, who identified the incorrect ARIA usage. We believe fails Web Content Accessibility Guidelines ([4.1.2 Name, Role, Value -  level A](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value))

### DIFF
--- a/lib/source/layouts/core.erb
+++ b/lib/source/layouts/core.erb
@@ -66,7 +66,7 @@
       <div class="app-pane__body"<%= " data-module=\"#{yield_content(:toc_module)}\"" if content_for? :toc_module %>>
         <% if content_for? :sidebar %>
           <div class="app-pane__toc">
-            <div class="toc" data-module="table-of-contents" tabindex="-1" aria-label="Table of contents">
+            <div class="toc" data-module="table-of-contents" tabindex="-1">
               <%= partial "layouts/search" %>
               <button type="button" class="toc__close js-toc-close" aria-controls="toc" aria-label="Hide table of contents"></button>
               <nav id="toc" class="js-toc-list toc__list" aria-labelledby="toc-heading"<%= " data-module=\"collapsible-navigation\"" if config[:tech_docs][:collapsible_nav] %>>
@@ -76,7 +76,7 @@
           </div>
         <% end %>
 
-        <div class="app-pane__content toc-open-disabled" aria-label="Content">
+        <div class="app-pane__content toc-open-disabled">
           <main id="content" class="technical-documentation" data-module="anchored-headings">
             <%= yield %>
             <%= partial "layouts/page_review" %>


tombye (approved):
Thanks for this Andrew. It is odd these `<div>`s having labels but no role and that's reflected in the experience using a screen reader. This makes it better and, actually, doesn't lose any semantics because the table of contents is already wrapped in a `<nav>` and the content a `<main>`, both of which do a good job of describing what each contain.

Nice work, only adds value 👍🏻

PR #406, opened 2025-06-05
Add aria-current to active header menu link

### LINKED ISSUE #403: Bug: Currently active menu item is not conveyed programmatically in the header
When documentation includes links in the header, for example on the [WCAG Primer page](https://alphagov.github.io/wcag-primer/index.html), the currently active menu item, such as Overview or All Requirements, is conveyed visually (by changing the text from white to blue) but not programmatically.

While it meets [WCAG 1.4.1 Use of Colour](https://www.w3.org/WAI/WCAG22/Understanding/use-of-color) (because the contrast ratio between white and blue is at least 3:1) it fails [WCAG 1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships) because information is not conveyed programmatically - for example by using `aria-current`.

Menu with most items in white and All Requirements in blue

This was originally raised as [WCAG Primer issue 132](https://github.com/alphagov/wcag-primer/issues/132).

andrewhick:
## What’s changed

This PR adds `aria-current="page"` to the currently highlighted menu item in the page header. I believe this fixes #403, an accessibility issue where the currently active link was only conveyed using a change of colour.

I've tested this on the example site using VoiceOver on Safari and Chrome. The currently active page reads correctly as "current page", and any non-current pages do not include this label.

## Identifying a user need

This came out of accessibility testing in the WCAG Primer working group by Shabana Ali. As the information is only conveyed visually and not programmatically, we believe it currently fails WCAG ([1.3.1 Info and Relationships - level A](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships)).

### DIFF
--- a/lib/source/layouts/_header.erb
+++ b/lib/source/layouts/_header.erb
@@ -40,7 +40,7 @@
         <ul id="navigation" class="govuk-header__navigation-list">
           <% config[:tech_docs][:header_links].each do |title, path| %>
             <li class="govuk-header__navigation-item<% if active_page(path) %> govuk-header__navigation-item--active<% end %>">
-              <a class="govuk-header__link" href="<%= get_path_to_resource config, path, current_page %>"><%= title %></a>
+              <a class="govuk-header__link" href="<%= get_path_to_resource config, path, current_page %>"<% if active_page(path) %> aria-current="page"<% end %>><%= title %></a>
             </li>
           <% end %>
         </ul>


tombye (approved):
Thanks for this Andrew. Works as expected when tested with the WCAG primer (Notify doesn't link to docs pages in our header so I can't test with our docs).

It also touches code that will just be the same whatever flavour of docs you have so good to go 👍🏻

Issue #402, opened 2025-05-30
Bug: Anchor links are not keyboard or screen reader accessible


andrewhick:
The link icon which appears next to each heading on hover is not keyboard accessible, or presented to screen reader users.

Options to fix:

1. Remove the anchor link functionality entirely from the gem - possibly drastic and we don't know how many people currently rely on this functionality.
2. Develop a configuration option to turn anchor links on or off - probably the most balanced option
3. Make the anchor links fully keyboard and screen reader accessible - the ideal option, but potentially difficult to implement

We believe this fails [WCAG 2.1.1 Keyboard](https://www.w3.org/WAI/WCAG22/Understanding/keyboard) and probably [4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value).

For context, this was originally raised as [WCAG Primer issue 131](https://github.com/alphagov/wcag-primer/issues/131) from accessibility testing on the Primer. Within the Primer group, we decided that although the functionality is potentially a useful enhancement for some people, it would be hard to make accessible and the easiest course of action is to remove it if we can. For comparison, GOV.UK does not have this functionality.

![A level 1 heading with a link icon along with its code. The link's code includes `aria-hidden="true" and tabindex="-1"](https://github.com/user-attachments/assets/03cae679-139e-4374-841c-0ea25fce4073)

PR #407, opened 2025-06-05
Add toggle for anchored headings

andrewhick:
## What’s changed

This PR adds a toggle option for anchored headings, where a hoverable link appears next to each heading in the main page body. It uses the same syntax as the `collapsible_nav` functionality.

As these links are not currently keyboard or screen reader accessible, the option to remove them is needed for a repository to meet Web Content Accessibility Guidelines. The [WCAG Primer](https://github.com/alphagov/wcag-primer) working group has also expressed a desire to turn these links off to reduce unnecessary page "noise".

While this does not completely fix issue #402 by making the links accessible (which would be a bigger fix), it does provide a workaround for it, as well as adding extra flexibility to the template.

I've tested it locally and it works as expected (you need to restart Middleman when changing the toggle). The default value for the new toggle is set to `true`, to reflect the behaviour that all Tech Docs repositories currently have.

## Identifying a user need

This came out of accessibility testing by Shabana Ali as part of the WCAG Primer working group. As the anchor links are not keyboard or screen reader accessible, we believe they fail WCAG ([2.1.2 Keyboard - level A](https://www.w3.org/WAI/WCAG22/Understanding/keyboard), and [4.1.2 Name, Role Value - level A](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value)).

### DIFF
--- a/example/config/tech-docs.yml
+++ b/example/config/tech-docs.yml
@@ -40,6 +40,9 @@ google_site_verification: dstbao8TVS^DRVDS&rv76
 
 enable_search: true
 
+# Enable hoverable anchor links next to every heading
+enable_anchored_headings: true
+
 show_contribution_banner: true
 
 github_repo: alphagov/example-repo
--- a/lib/source/layouts/core.erb
+++ b/lib/source/layouts/core.erb
@@ -77,7 +77,7 @@
         <% end %>
 
         <div class="app-pane__content toc-open-disabled" aria-label="Content">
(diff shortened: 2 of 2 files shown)


andrewhick:
Thanks @tombye , I've tested this locally and it works as expected for me. The anchor links:

- appear when the toggle is `true`
- do not appear when the toggle is `false`
- appear when the toggle is not present

tombye (approved):
Looking good now, thanks for making those changes 👍🏻

PR #415, opened 2025-06-16
Apply brand refresh to tech docs using `show_govuk_logo`

owenatgov:
Full description coming very soon

# What’s changed


spec/features/integration_spec.rb
```diff
@@ -136,4 +145,16 @@ def when_i_view_a_page_with_prevent_indexing
...
+    expect(page).to have_css ".govuk-template--rebranded"
+  end
+
+  def then_the_header_is_using_the_new_logo
+    expect(page).to have_css ".govuk-logo-dot"
+  end
+
+  def then_the_footer_is_showing_the_crown
+    expect(page).to have_css ".govuk-footer__crown"
+  end
```
romaricpascal, line 159: **nitpick** Thinking we could have a  single `then_the_brand_refresh_is_enabled` which checks all three, similarly to how `and_there_are_proper_meta_tags` does multiple expectations. This'll make the test more succint and we won't lose which part of the test fail as we should know which `expect` threw from the stacktrace when the test fails.

romaricpascal (approved):
Looking grand, only comment is a little nitpick to make the tests a little more succint 😊

PR #418, opened 2025-06-23
Update logo in header

romaricpascal:
GOV.UK Frontend has updated the SVG for the logo to fix artifacts at high zoom levels:
https://github.com/alphagov/govuk-frontend/pull/6036

This PR updates the header accordingly (using the markup from the macro as the CHANGELOG had some discrepancy)


owenatgov (commented):
That do be the correct path. Changelog update pls.

owenatgov (approved):
:shipit:

PR #419, opened 2025-06-23
Update path to GOV.UK Frontend assets when using GOV.UK logo

romaricpascal:
These were pointing to the locations before the brand refresh, so the wrong favicon were showing. Includes the favicons in the header, as well as the Open Graph image.

Paths to the assets when not using the GOV.UK logo were left untouched, as they are built from separate options.

In action:
Screenshot of the Tech Docs gem with the right paths for the favicons


owenatgov (approved):
:shipit:

PR #423, opened 2025-06-24
Fix copy of `manifest.json` into build

romaricpascal:
Uses a glob to target any JSON file within the `assets` folder of GOV.UK Frontend so that future changes of manifest location get copied properly.

This needs to be a separate glob from the initial list to not copy other JSON files in the package, like `package.json` and similar. Only the assets need copying.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+### Fixes
+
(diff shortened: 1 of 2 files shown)


owenatgov (approved):
I've tested this on a local instance and everything is working as expected. This just needs a changelog entry and we're grand.

PR #434, opened 2025-09-17
Release v5.1.1

tombye:
Patch release to change the method of installation npm uses. Includes https://github.com/alphagov/tech-docs-gem/pull/433


kr8n3r (approved):
nice one

PR #187, opened 2020-07-20
Google Tag Manager layout addition

grahampaulcook:
Introduce Google Tag Manager support into technical documentation services by adding a new configuration option in the `config/tech-docs.yml` file.

`gtm_id: GTM-xxxxx`

PR #436, opened 2025-11-14
[PP-7355] Support Google Tag Manager

yndajas:
## What’s changed

This adds support for the Google Tag Manager (GTM) approach to Google Analytics 4

Previously #350 added Google Analytics 4 support, but though it mentioned Tag Manager, if my understanding is correct it actually added support for the non-GTM approach: gtag.js. See:
- https://developers.google.com/tag-platform/tag-manager/datalayer
- https://support.google.com/tagmanager/answer/7582054

There is already a PR open to add GTM support (#187). This differs slightly: it adds a cookie banner consent step into the flow when using GTM, and doesn't include the no script (JavaScript-less) tracking (to avoid adding complexity to the consent process)

## Identifying a user need

We need to use this on https://github.com/alphagov/govuk-content-api-docs to track usage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### New features
+
+- [Support using Google Tag Manager for Google Analytics 4](https://github.com/alphagov/tech-docs-gem/pull/436)
+
 ## 5.1.1
 
 ### Fixes


lib/source/layouts/core.erb
```diff
@@ -52,6 +52,28 @@
...
+          <div class="govuk-button-group">
+            <button type="button" class="govuk-button" data-module="govuk-button">
+              Accept analytics cookies
+            </button>
+            <button type="button" class="govuk-button" data-module="govuk-button">
+              Reject analytics cookies
+            </button>
+          </div>
+        </div>
+      </div>
```
richardTowers: It might be a bit easier to read if you moved this into a partial.

yndajas:
> I think you'll also need to add this to the CHANGELOG and bump the version (unless you're planning on doing that in a separate PR?)

I've updated the CHANGELOG but it looks like the preference is to bump the version in a separate release PR in this repo: https://github.com/alphagov/tech-docs-gem#releasing-new-versions

richardTowers (approved):
Thanks Ynda - looks good to me.

I think given this project is (at best) community maintained, and I've dipped in and out of maintaining it, my approval should be enough for a merge. Barring anyone else popping up anyway!

PR #438, opened 2026-02-06
Fix `select_top_level_html_files` ignoring custom `http_prefix`

inulty-dfe:
## What’s changed

When we configure a custom `http_prefix` (e.g. /docs) to serve our tech docs site under a subpath, Middleman sets the parent URL of top-level pages to `/docs/` rather than `/`.

`select_top_level_html_files` had the parent URL check hardcoded to "/", so it filtered out all our top-level pages — meaning the multi-page table of contents came back empty.

`render_page_tree` (in the same file) already handled this correctly by computing `home_url` from `config[:http_prefix]`. `select_top_level_html_files` just wasn't updated to match.

[Here is where we've had to monkey patch the library](https://github.com/DFE-Digital/publish-teacher-training/blob/main/docs/config.rb#L6-L14)

### DIFF
--- a/lib/govuk_tech_docs/table_of_contents/helpers.rb
+++ b/lib/govuk_tech_docs/table_of_contents/helpers.rb
@@ -30,14 +30,16 @@ module GovukTechDocs
           .to_a
       end
 
-      def select_top_level_html_files(resources)
+      def select_top_level_html_files(resources, config)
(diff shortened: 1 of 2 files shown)


JC-DSIT:
no response from review.  Closing this PR and linking to  issue https://github.com/alphagov/tech-docs-gem/issues/488

PR #439, opened 2026-02-17
add option to skip rendering child when rendering ToC

JC-DSIT:
## What’s changed

We have added an optional flag for the `multi_page_table_of_contents/render_page_tree` helper functions.  The table of contents (ToC) currently has limitations when using layouts to group sections and create filtered ToC - e.g. when using `header_links` and requiring a specific ToC .  

### Example usage

```
<% section_index_page = sitemap.resources.find {|resource| resource.path == "section/index.html" } %>
<% content_for :sidebar do %>
   <%= render_page_tree [section_index_page], current_page, config, yield, include_child_resources: false %>
<% end %>

<% wrap_layout :core do %>
  <%= yield %>
  <% content_for(:toc_module, "in-page-navigation") %>
<% end %>

```

## Identifying a user need

This is particularly an issue when we have a directory with an index page, and then sub directories containing nested sections.  When trying to organise this structure using `header_links` to create sections, we want each section to create it's own ToC  eg:

```
source/
├── index.html                   
├── contact.html
├── section/ # this is the link used in the `header_links` href
  └── index.html
       ├── sub-section/
           ├── index.html 
           ├── subpage1.html     
           ├── subpage2.html
```

The expected result would be:

desired result

This bug has been [previously raised here](https://github.com/alphagov/tech-docs-template/issues/225).  We are currently unable to publish updated public facing documentation due to the navigation being too painful.  This fix allows us to do this, while maintaining backwards compatibility

#### Using resource.children

If we use the resource.children call then we loose the section index page altogether.  This page can only be found by clicking the link in the navigation (top right).  This is a confusing user journey and very unclear.

code example:

```
<% section_index_page = sitemap.resources.find {|resource| resource.path == "section/index.html" } %> # using a parent resource creates issues similar to those outlined below, and doesn't solve the problem
....
 <%= render_page_tree section_index_page.children, current_page, config, yield %>
```

just doing children

#### Including resource in tree builder

If we try to include the the extra resource then the full child tree is rendered below the main section.  Result below is the same if we merge into a single array etc

code example:

```
<%= render_page_tree [section_index_page], current_page, config, yield %>
 <%= render_page_tree section_index_page.children, current_page, config, yield %>
```

result:

duplicate full tree

#### Using single page ToC

When using the single page ToC functions the ToC changes when you move to a different page, with the section index missing altogether

```
<%= single_page_table_of_contents(yield) %>
<%= render_page_tree section_index_page.children, current_page, config, yield %>
```

single page toc

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,51 +17,60 @@ This gem uses [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend), par
 
 We use `npm` to download the govuk-frontend package. To update to a new version, change the version in the [package.json file](package.json) and run `npm update`.
 
-## Use layouts to structure your Table of Contents
+## Table of contents helper functions
 
-You can use layouts to group pages and customise your Table of Contents (ToC).  To create a template create a new ruby file in the `source/layouts` directory, for example `source/layouts/live_support_layout.erb`.  You can apply the template by including it in the frontmatter block of your required pages, for example:
-
-```diff
-    ---
-    title: Live support rota
-    weight: 1
-    last_reviewed_on: 1970-01-01
-    review_in: 3 months
-+   layout: live_support_layout 
-    ---
-```
-
-
-This gem has helper functions to render your ToC.  You should use:
+With `Middleman` you can apply layouts to group pages and customise sites. This gem has the following additional helper functions to manage the table of contents (ToC):
 
 - `single_page_table_of_contents` to create a ToC from the headings on the current page
 - `multi_page_table_of_contents` to create a ToC for a group of pages, opened at the current page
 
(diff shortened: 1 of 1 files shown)


claire-cheuk (dismissed):

README.md
```diff
@@ -17,6 +17,52 @@ This gem uses [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend), par
...
+    title: Live support rota
+    weight: 1
+    last_reviewed_on: 1970-01-01
+    review_in: 3 months
++   layout: live_support_layout 
+    ---
+```
+
+
+This gem has helper functions to render your ToC.  You should use:
```
NathanD-GDS:
```suggestion
This gem has helper functions to render your ToC.  You can use:
```

Is it 'should' or 'can'?

PR #442, opened 2026-03-13
Bump govuk frontend to v6

### LINKED ISSUE #431: Use new Gov UK branding
Use new Gov UK branding with blue banner

tombye:
## What’s changed

This bumps the version of Design System (GOV.UK Frontend) used by this gem to V6 and packages it up as a beta release, to allow downstream projects time to try it out.

example_multi-page_new_look

## Identifying a user need

Public documentation should have the GOV.UK brand, so users get confirmation they're looking at the right thing because it matches both the service and www.gov.uk. Non-public documentation should be consistent with other docs, as much as possible, for similar reasons.

## Notes for reviewers

Because the new branding style, which moves the header links to a service navigation component below the GOV.UK logo, is now mandatory, I'm calling this a major version bump.

Other than those changes, the other things changed are:
- the introduction of a new colour palette
- the removal of the 14px font size
- the footer links being changed to match all other links in the page

### Examples of downstream projects with this version of the gem
Based on the [spreadsheet of tech docs using the tech docs gem across GDS](https://docs.google.com/spreadsheets/d/1LxO3E972uEU6muXevpzHDb2vRyz5dh4vfPs3QzdQwY0/edit?usp=sharing)

#### GDS Way

gds_way_new_look

#### WCAG Primer

wcag_primer_new_look


kr8n3r (approved):
this is a marked as a major version so dependency update tools will not auto-upgrade it.
Teams that use this this gem will be able to evaluate, test and make changes in their own timeframe

PR #443, opened 2026-03-17
Add option to use GOV.UK Design System components

### LINKED ISSUE #173: Allow warning_text helper to accept nested content (blocks)
We can use the `warning_text` help with a plain sentence to output a warning: 

https://tdt-documentation.london.cloudapps.digital/amend_project/content/#insert-a-warning

If we want to put more complicated markup into this helper it does not work, for example a link.

We should consider making this helper work with blocks, for example.

```erb
<%= warning_text do %>
    <%= markdown(doc.deprecated) %>
<% end %>
```

### LINKED ISSUE #458: Release 6.1.0.beta
###  [Bugfix for active tabs](https://github.com/alphagov/tech-docs-gem/pull/452)
This bug fix:

- fixes an issue where some a nested structure can cause `header_links`  to be incorrectly set or unset as 'active'
- adds a new test suite to clearly test and define the expected behaviour of the active tab

### [Enable GOV.UK Design System components](https://github.com/alphagov/tech-docs-gem/pull/443)

This enhancement:
- adds a `nunjucks` render extension to allow calling and rendering of GOV.UK components
- adds an option configuration to enable this extension

JC-DSIT:
## What’s changed

I have added a custom extension and renderer to load in gov.uk nunjucks components.  They will call these in the same way documented on the [design system](https://design-system.service.gov.uk/get-started/) documentation.

The extension will recogniser any function call that has the prefix `govuk` and pass the call and any data to the renderer.  If the template is not known then the relevant error will be passed back up the chain.

Users can disable this by setting adding 

`enable_govuk_components: false` in `config/tech-docs.yaml`.

## Identifying a user need

There are examples around existing tech docs of users creating their own hard coded copies of gov.uk extensions for example:

- [cookie banner](https://github.com/govuk-one-login/tech-docs/blob/e664dd48f959f54d5f5e9d34473533b954cb0362/source/partials/_cookie-banner.erb)
- [notification banner](https://github.com/govuk-one-login/team-manual/blob/c5e2051e74a18f8d32a5ec72964576635cd8fdb8/source/supporting-govuk-one-login/services-we-support/identity-proving/mobile-app-verification.html.md.erb#L10)
- [warning text](https://github.com/alphagov/tech-docs-gem/blob/main/lib/govuk_tech_docs/warning_text_extension.rb) (in the gem itself)

There are others, and likely ones that we don't have visibility of. 

These do not get updated with new releases of the design system, and need manul maintenance.  

This will expand the tech-docs-gem to enable using user and accessibility tested components, while discouraging custom html.  It also supports consistent gov.uk documentation, while enabling flexibility to create templates that better suit individual requirements.

And if nothing else, we say we do it [here](https://github.com/alphagov/tdt-documentation/blob/ece36a0700609c1c2b779f8e5cd4b5ba058867f5/source/write_docs/content/index.html.md.erb#L125) but at the moment, we don't...


JC-DSIT:
merged into release 6.1.0.beta branch - #458

PR #463, opened 2026-04-13
[PP-7656] Update `openapi3_parser` to 0.10.1

brucebolt:
## What’s changed

Updates the `openapi3_parser` dependency to the latest version.

## Identifying a user need

`openapi3_parser` 0.9.0 has a dependency on `commonmarker` 0.x, which does not support Ruby 4.

Updating to 0.10.x, which allows us to use `commonmarker > 1`, therefore unlocking the Ruby 4 upgrade in apps that use this gem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+- Update version of `openapi3_parser` allowing support for Ruby 4.0.
+
 ## 6.0.0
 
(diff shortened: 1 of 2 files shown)


brucebolt:
This should not be merged until https://github.com/alphagov/tech-docs-gem/pull/464 is merged.

yndajas (approved):
In case you run into any issues later: https://github.com/alphagov/content-store/pull/1598/changes/89bae5cb36f0f11afa34e911af75a07c8d4d8d54

PR #464, opened 2026-04-13
[PP-7656] Add Ruby 4.0 to testing matrix

brucebolt:
## What’s changed

Adds Ruby 4.0 to the CI test matrix.

## Identifying a user need

This version of Ruby has recently been released, so we should ensure the gem is tested against the latest version as apps using this gem may already be on Ruby 4.


JC-DSIT (approved):
LGTM

PR #465, opened 2026-04-14
- remove beta

JC-DSIT:
Removes beta tag on release to make version 6 official


brucebolt (approved):
Thanks for making this change.

I was attempting to release a patch of v5, to avoid using the beta version of this gem.  I'll now just upgrade to 6.x.

PR #467, opened 2026-04-14
Raise minimum Ruby version to 3.3

### LINKED ISSUE #469: Raise minimum Ruby version to 3.3
#467 

What’s changed

Remove support for Ruby 3.2.

Identifying a user need

Ruby 3.2 is now out of support, so we do not expect consumers of the gem to be using

brucebolt:
## What’s changed

Remove support for Ruby 3.2.

## Identifying a user need

Ruby 3.2 is now out of support, so we do not expect consumers of the gem to be using it.


.ruby-version
```diff
@@ -1 +1 @@
-3.2.6
+3.3.11
```
JC-DSIT, line 1: Just think it might be best to use the ame version everywhere for consistency simplicity, plus 3.3.11 was only released 3 weeks ago so might be a bit keen 😄 

```suggestion
3.3.0
```

JC-DSIT:
I've tested this locally and it builds and runs correctly, so looks good to me.  

Could do with confirming it's still compatible with the tech-docs-template

.ruby-version
brucebolt: Generally on GOV.UK (other parts of GDS may do this differently), we set `.ruby-version` to be the latest release of the minimum supported version.  We do that because the version is only used when running the tests locally.  Actual consumers of the gem can use any version of 3.3 they like.

JC-DSIT: yeah I get you, i just always prefer consistency within a codebase, so if the gemspec says `spec.required_ruby_version = ">= 3.3.0"` i'd normally just use the same throughout. Happy for you to accept or disregard the suggestion, the result is the same as you say 🙂

JC-DSIT:
@brucebolt - I have opened a PR for the tech-docs-template to up the ruby version there also https://github.com/alphagov/tech-docs-template/pull/248/changes

CHANGELOG.md
```diff
@@ -1,10 +1,11 @@
-# Unreleased changes
+## Unreleased
```
JC-DSIT: Do you want to do this as release 6.2.0?

brucebolt: Have pushed a commit to release 6.2.0 to this PR.

JC-DSIT (approved):

PR #477, opened 2026-04-28
Ensure comments in code examples are readable

NickColley:
### What changed

Fixes comments being set to a light grey which made them very difficult to read.

The contrast ratio with secondary-text against the background is 8.57:1 so we do not need to darken it and therefore can avoid the manual manipulation of the tint colours.

Downstream fix in the GOV.UK Frontend technical documentation:
https://github.com/alphagov/govuk-frontend-docs/pull/621

### Before
Screenshot of code example with comments that have very low contrast 

### After
Screenshot of code example with comments that have very low contrast

## Checklist

Before you request approval you should confirm that:

- [ ] GitHub actions all pass
- [ ] you have tested the changes with a fresh build against the latest version of the `tech-docs-gem`
- [ ] you have linked this PR to an issue (or explained why none exists)
- [x] you have updated documentation if needed


NickColley:
Note that I have not tested this locally as I do not know how to develop locally for this gem so while this should work I could use a hand testing this out on my machine 👍🏻

JC-DSIT:
I've tested this locally and it works.  
The specific issue comes from specifying `html` around the codeblock e.g:
``` 
`html `
```
This causes the highlighting to act differently.  The comments still appear for example if you just have the backticks, or if you choose a different wrapper (e.g shell)

Screenshot 2026-04-29 at 09 55 49

gives 

Screenshot 2026-04-29 at 09 56 25

It raises some concerns though:

- is that a design system choice.  I.e. does the design system account for the myriad options one can put around backticks?
- if they don't are we confident the colours are accessibility tested?
- does the suggested change break any other code wrapping styles?

I'm happy to add this in for now, but we will need to follow up.

@NickColley need to up the version and update the changelog before approval 😄

NickColley:
> is that a design system choice.  I.e. does the design system account for the myriad options one can put around backticks?

Syntax highlighting like this is wide spread used across Tech Docs Gem usage

> does the suggested change break any other code wrapping styles?

This change reverts the colour to the previous colour before it was broken, so essentially it is the same as it was before.


> if they don't are we confident the colours are accessibility tested?

I believe the colours when they were updated in the previous work were tested but this could be double checked separately to this pull request.

I have tested the contrast for this change as described in this pull request, if there's worries about colour contrast issues broader an issue should be raised to investigate that but we can go ahead and fix this bug while that happens.

JC-DSIT (approved):

PR #480, opened 2026-05-01
Fix TOC sidebar not scrolling to active item on page load

### LINKED ISSUE #479: [Bug]: TOC scroll to active item
### tech-docs-gem version

6.2.1

### ruby version

_No response_

### What happened?

When selecting a TOC item, the TOC item does not always remain in view in the sidebar when the new page is loaded.

### Steps to reproduce

can be reproduced by visiting the Register Trainee Teachers API documentation.

1. Go to https://www.register-trainee-teachers.service.gov.uk/csv-docs/
2. expand "v2025.0"
3. click on the first item at the top of the list
4. The TOC item will not be visible in the sidebar (user must scroll to see it/find it)

### Logs or error output

_No response_

### Checks

- [x] I have searched for existing issues
- [x] I have checked other users also have this issue
- [x] I have checked that I am not using a local or forked repository with untested changes

d-a-v-e:
### What changed

`scrollTocToActiveItem()` in `_modules/in-page-navigation.js` used jQuery's `position().top` to compute whether the active TOC item needed scrolling into view. 

When called on initial page load (via `handleInitialLoadEvent`), the browser has not yet finished layout, so `position().top` returns 0 and the function silently exits without scrolling.

Two fixes:
- Wrap the initial-load call in `requestAnimationFrame` so layout is complete  before positions are read.
- Replace the manual offset calculation with the browser-native `scrollIntoView({ block: 'nearest', inline: 'nearest' })`, which is simpler and handles all edge cases.

Also:
-  converts var to let/const throughout the file (as flagged by the linter)

- Fixes  #479 

### Before

https://github.com/user-attachments/assets/15aca9b0-7e4e-4cee-9e5f-62b7a335e56e

### After

https://github.com/user-attachments/assets/c0e0f76e-4e94-4711-b079-04d85a376d48

## Checklist

- [x] GitHub actions all pass
- [x] you have tested the changes with a fresh build against the latest version of the `tech-docs-gem`
- [x] you have linked this PR to an issue (or explained why none exists)
- [ ] you have updated documentation if needed

### DIFF
--- a/lib/assets/javascripts/_modules/in-page-navigation.js
+++ b/lib/assets/javascripts/_modules/in-page-navigation.js
@@ -2,13 +2,11 @@
   'use strict'
 
   Modules.InPageNavigation = function InPageNavigation () {
-    var $tocPane
-    var $contentPane
-    var $tocItems
-    var $targets
+    let $contentPane
+    let $tocItems
+    let $targets
 
     this.start = function start ($element) {
-      $tocPane = $element.find('.app-pane__toc')
       $contentPane = $element.find('.app-pane__content')
       $tocItems = $('.js-toc-list').find('a')
       $targets = $contentPane.find('[id]')
@@ -29,7 +27,7 @@
         } else {
           // Store the initial position so that we can restore it even if we
           // never scroll.
-          handleInitialLoadEvent()
+          window.requestAnimationFrame(handleInitialLoadEvent)
         }
       }
     }
@@ -41,7 +39,7 @@
     }
 
     function handleInitialLoadEvent () {
-      var fragment = fragmentForTargetElement()
+      let fragment = fragmentForTargetElement()
 
       if (!fragment) {
         fragment = fragmentForFirstElementInView()
@@ -51,7 +49,7 @@
     }
 
     function handleScrollEvent () {
-      var fragment = fragmentForFirstElementInView()
+      const fragment = fragmentForFirstElementInView()
 
       storeCurrentPositionInHistoryApi(fragment)
       highlightActiveItemInToc(fragment)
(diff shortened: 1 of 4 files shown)


lib/assets/javascripts/_modules/in-page-navigation.js
```diff
@@ -91,38 +89,25 @@
...
-
-      var offset = null
-
-      if (linkTop < 0) {
-        offset = linkTop
-      } else if (linkBottom >= paneHeight) {
-        offset = -(paneHeight - linkBottom)
-      } else {
-        return
+      const el = $activeTocItem.get(0)
```
JC-DSIT: Since line 84 does the defensive check you can simplify this:

```diff
//lne 84 - 88

if ($activeTocItem.get(0)) {
        $tocItems.removeClass('toc-link--in-view')
        $activeTocItem.addClass('toc-link--in-view')
-        scrollTocToActiveItem($activeTocItem)
+.       scrollTocToActiveItem($activeTocItem.get(0))
      }

// line 92 (variable names should ideally describe what they are, "el" isn't super clear)
      function scrollTocToActiveItem(activeTocElement){ 
-       const activeTocElement = $activeTocItem.get(0)
+       if (activeTocElement && typeof el.scrollIntoView === 'function') .....
      
```
If you want to make this function to not trust what it's given (which is fair) you should add the defensive check here too i.e. not assume `$activeTocItem.get(0)` exists

JC-DSIT:
I don't think this fully solves the issues with the toc for all the ways the tech-docs can be structured, for example when using layouts to group items under the navigation menu.   

It looks fine and I have tested on a couple of sites locally which all still run and work well, have added a small comment but would ideally like to see [some testing added to prevent future breakages ](https://github.com/alphagov/tech-docs-gem/blob/main/spec/javascripts/table-of-contents-spec.js)

You'll also need to up the release version and update the changelog

d-a-v-e:
thanks for the Review @JC-DSIT - I've added some specs, do let me know if this looks sufficient and I can look at bumping versions + changelog (I assume in a separate PR?)

JC-DSIT:
@d-a-v-e looks fine to me - no need to add the up version and changelog in a different PR I don't think?

I'll set the checks to run now and as long as they all pass then once there's a new version it should be fine to merge in 👍

JC-DSIT (approved):
Nice, looks good to me

PR #513, opened 2026-06-15
[Bug]: whitespace on long pages

JC-DSIT:
## Proposed changes

The main page layout in the `tech-docs-gem` is not laid out correctly.  This means that the footer does not fill the space correctly:

Screenshot 2026-06-15 at 10 48 52

### What changed

- re ordered the template elements so they stack correctly
- removed `.app-pane` from wrapper as this introduces an in-page overflow scroll

Result:

Screenshot 2026-06-15 at 10 54 13

Tested using Google Dev tools responsive mode and looks unaffected

## Related issue or tracking reference

You should have an open GitHub issue that this PR will fix.  

- Fixes #
- Relates to [template issue 258](https://github.com/alphagov/tech-docs-template/issues/258)

> If there is no issue, please open one or please explain why one is not needed (for example small maintenance change, routine dependency update).

## Checklist

Before you request approval you should confirm that:

- [x] the pull request has a clear title with a short description about the update in the documentation
- [x] GitHub actions all pass
- [x] you have tested the changes with a fresh build against the latest version of the `tech-docs-gem`
- [x] you have linked this PR to an issue (or explained why none exists)
- [x] you have updated documentation if needed

PR #538, opened 2026-09-13
Fix sticky TOC scrolling after app-pane wrapper removal

### LINKED ISSUE #537: [Enhancement]: Fix scrolling behaviour so navigation scrolls independently of content
### What should change?

A change was made in [PR513](https://github.com/alphagov/tech-docs-gem/pull/513) which prevents the table of contents from scrolling independently of the content. This makes the scroll position of the TOC jump around when moving between sections. Ideally they should scroll independently of one another.

### User need

Help the user stay oriented in the TOC by keep it in position when they're scrolling through the content. 

### Proposed solution

_No response_

gonogo:
Restore an independently scrollable sidebar without locking the viewport, so the full-width footer scrolls into view with the page.

## Proposed changes

- Restore an independently scrollable left TOC after the `.app-pane` wrapper was removed in #513, without locking the viewport.
- Keep the full-width footer in normal document flow so it scrolls into view at the end of the content.
- Update in-page and collapsible navigation scroll handling to listen on `window` instead of `.app-pane__content`.

- Relates to #513
- Fixes #537


JC-DSIT:
@gonogo - thanks for the PR, a couple of issues I noticed in the screen grab below :) 

https://github.com/user-attachments/assets/85a5a139-40d9-4485-b705-c22a2711625f

JC-DSIT (changes requested):
See screen recording attached to PR comments.
- border is not consistent
- screen bounces on long ToC

gonogo:
Good spot, have now fixed footer issue. Not sure what you mean about inconsistent border though? Can you clarify? Thanks

JC-DSIT:
> Good spot, have now fixed footer issue. Not sure what you mean about inconsistent border though? Can you clarify? Thanks

Thanks @gonogo - regarding the border these screenshots may be clearer:

Toc Prod
Toc Error 2
ToC error 1


Either the right border of the ToC container needs to be visible for the full length at all times, or the left border of the main content.  It doesn't really matter which so long as the result is the same, so whichever is easier/most sensible

PR #540, opened 2026-09-15
Read report issue URL from tech docs configuration

### LINKED ISSUE #283: `report_issue_url` cannot be configured in `config/tech-docs.yaml`
Currently the tech docs documentation tells users that they can configure the url for the 'report problem' link by adding some configuration to `config/tech-docs.yaml` [[1]]. Unfortunately, a bug in the implementation of this feature means that the configuration actually has to be added to `config.rb`:

https://github.com/alphagov/tech-docs-gem/blob/66f71178f2956f9acbc70ef182a786f12768ff76/lib/govuk_tech_docs/contribution_banner.rb#L22

Line 22 should actually read `source_urls` from the `tech_docs` hash.

[1]: https://tdt-documentation.london.cloudapps.digital/configure_project/global_configuration/#source-urls

## What should change

We should change it so that the configuration is read from `config/tech-docs.yaml`, while retaining a fallback ability to configure from `config.rb` as before, to avoid a breaking change.

## User need

As a user I need to be able to use example code from the documentation.

zaid3:
## What changed

- Read `report_issue_url` from the `source_urls` section of `config/tech-docs.yml`.
- Preserve the existing `config.rb` location as a backwards-compatible fallback.
- Add regression coverage for both configuration paths.

## Why

The documented configuration was not being read from `config/tech-docs.yml`, making the example configuration ineffective.

## Issue

Fixes #283

## Testing

- Added tests for the documented YAML configuration.
- Preserved and tested the legacy configuration fallback.
- Automated tests will run through the repository’s GitHub Actions workflow.


JC-DSIT (approved):

JC-DSIT (approved):
@zaid3 - happy to merge this into the release branch once the jobs are able to run.  At the moment it needs `npm audit fix` to patch some npm issues.

zaid3:
Thanks, @JC-DSIT . I have updated the lockfile to move js-yaml from 4.3.1 to 4.3.2. The npm audit now reports zero vulnerabilities. The updated commit is ready for the checks to be run.

JC-DSIT:
> Thanks, @JC-DSIT . I have updated the lockfile to move js-yaml from 4.3.1 to 4.3.2. The npm audit now reports zero vulnerabilities. The updated commit is ready for the checks to be run.

@zaid3 no worries thanks for fixing the audit

PR #541, opened 2026-09-15
Change default branch from `master` to `main` in contribution banner

connorgurney:
### What changed

I've changed the default branch in repository URLs in the contribution banner from `master` to `main`.

This removes the need for GitHub to redirect from the former to the latter where repositories have already migrated to the latter. It's a minor improvement but an improvement all the same.

I've not opened an issue as the actual change is so small. However, it might need one given how breaking it's likely to be, so I'd appreciate guidance.

## Screenshots or examples (if relevant)

Not applicable

## Checklist

Before you request approval you should confirm that:

- [ ] GitHub actions all pass
- [x] you have tested the changes with a fresh build against the latest version of the `tech-docs-gem`
- [x] you have linked this PR to an issue (or explained why none exists)
- [x] you have updated documentation if needed


JC-DSIT (changes requested):
Hi @connorgurney  thanks for contributing 😄 

I'm giving this some thought as to whether the breaking change is worth the reward.  The documentation says:

>Note: When creating a project using the template this is set to “main”, but if this setting is not present the default is currently “master”. This will change in a future major release.

So most projects presumably get it [set to main here](https://github.com/alphagov/tech-docs-template/blob/main/optional/config/tech-docs.yml.tt#L37):

```
show_contribution_banner: true
github_repo: alphagov/YOUR_REPO
github_branch: main
```

So if people don't want it to be main, they change it to whatever.  If they don't change it it's still main, if they remove it altogether it's set to master which presumably works for them (they've have to actively remove it).

Although I see the reasoning for the change and support it in theory, at this point where the gem is used by over 300 projects across gov, is it worth creating a job for people and/or risking causing issues for users just so that github saves a redirect?

Interested to hear your thoughts 🙂 

(p.s. If you do think so, then you need to update the changelog and update/add any tests for this function)

connorgurney:
Thanks, JC - appreciate you taking the time to review this.

I was unaware of the scale at which the gem was used, but, knowing that now, I think you're right. Whilst it'd be nice to have, I think that's the extent of it, so it doesn't seem to me that it justifies the amount of work for such limited gain.

I'll wait for your feedback before closing the PR, but thanks for reviewing it all the same - it's all part of the journey!

JC-DSIT:
> Thanks, JC - appreciate you taking the time to review this.
> 
> I was unaware of the scale at which the gem was used, but, knowing that now, I think you're right. Whilst it'd be nice to have, I think that's the extent of it, so it doesn't seem to me that it justifies the amount of work for such limited gain.
> 
> I'll wait for your feedback before closing the PR, but thanks for reviewing it all the same - it's all part of the journey!

Thanks @connorgurney - I'll close this for now then.  Please do open an issue any time you feel something could/should be improved (no matter how small!) as we have a project board  that pulls in open issues and helps us with planning, prioritising and creating releases - we welcome as many contributors as possible so don't be shy!  Also there are a bunch of open issues on the repo, so if you're looking to get involved and expand your knowledge of the gem please feel free to pick any of them up 🙂

